PR #306, opened 2017-11-06
new features for ASTi

JoshData:
The first commit makes it possible to open apps/modules in a new tab from the app page.

The second two commits are features that helped me improve the ASTi SSP.


JoshData:
This PR now has additional bug fixes and small UI improvements.

PR #311, opened 2017-11-07
Use curl instead of wget in fetch-vendor-resources.sh

peterkaminski:
On minimal systems (and macOS), it's probably more likely that `curl` will be present rather than `wget`.  Also, the foreign `google-font-download` script uses `curl` already.

This pull request is to replace the `wget` calls in `fetch-vendor-resources.sh` with `curl` calls.  FYI, curl needs the -L flag to follow redirects, which is required for at least some of the downloads.

Also since the file was open, I typo-fixed a comment, and added `rm -f /tmp/google-font-download` at the bottom to clean that file out of the /tmp directory.

Background: I was running the script on a minimal Ubuntu 16.04 LTS image; neither curl nor wget was installed.  It's not hard to install both, but curl is often already installed on systems, and being conservative with dependencies is generally a good thing too.

### DIFF
--- a/fetch-vendor-resources.sh
+++ b/fetch-vendor-resources.sh
@@ -68,7 +68,7 @@ curl -L -o $VENDOR/js-yaml.min.js \
 
 # google fonts
 #  Hind: SIL Open Font License 1.1
-# first download a helper (note: we're about to run a foriegn script locally)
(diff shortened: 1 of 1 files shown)


JoshData:
LGTM! Merging.

Issue #287, opened 2017-09-13
ensure all pip packages have version numbers


JoshData:
Packages without explicit version numbers in requirements.txt aren't automatically updated by pip without --upgrade. We should ensure every package has a version.

ad061d8556ee0ed9032ea47870c5543cf793b120

JoshData:
This is done. All of our immediate dependencies are now pinned to versions. We can remove `--upgrade` from our update scripts.

But the dependencies of our dependencies (the transitive dependencies, as Greg says) are not all pinned.

We also should add `--require-hashes` to `pip install` and place hashes in our `requirements.txt` file. This requires that all of our transitive dependencies also be listed in our requirements file with their hashes, and it requires that the transitive dependencies all be pinned too.

JoshData:
This was fixed in 88e4d2baa2874df1c1b2b946ddba720109535666.

Issue #314, opened 2017-11-07
reminder: fetch-vendor-resources.sh security and dependencies


peterkaminski:
This issue is just a reminder to be mindful that `/fetch-vendor-resources.sh` pulls in components and utilities with hardcoded URLs and versions.  It does so with straightforward code, but not with a standard library handler, so automated security and code quality may either flag it, or miss it, in which case we should make sure it gets reviewed properly some other way.

Also, as noted in one of the comments, it downloads and executes a foreign script, from `https://raw.githubusercontent.com/neverpanic/google-font-download/master/google-font-download`.

JoshData:
In another project I have a bash function [wget_verify](https://github.com/mail-in-a-box/mailinabox/blob/master/setup/functions.sh#L174) that downloads a file and checks its hash against a hard-coded value before continuing. We could use that. It'll go along with the work I've been doing today on adding hashes to our requirements.txt file.

JoshData:
We're now checking downloads against hard-coded hashes in the script.

@gregelin wrote in #321:

> The fetch-vendor-resources.sh which checks the hashes of downloaded assets like Bootstrap css and Google fonts contains hardcoded hashes. It would be better if these were in a configuration file like how we started handling Python packages in merge #318. However, it is not a critical short coming at this point.

The general idea of what's left in this issue is turning the script into something that statically specifies the downloads & hashes, rather than it being a bash script.

Issue #209, opened 2016-12-14
vendorize Emojione


JoshData:
Bring the Emojione files in locally and serve them statically rather than using a CDN.

from #208

PR #318, opened 2017-11-08
Ensure all dependencies are pinned and hashed

JoshData:
This PR ensures all of our dependencies are cryptographically verified and, for Python packages, checked against a database of known vulnerable package versions.
    
* Renamed requirements.txt to requirements.in, where we'll continue to manually store our (immediate) dependencies, and un-pinned some packages there so we more easily track the latest version.
* Made a new requirements.txt that has all of our (transitive) dependencies, pinned to versions, and with hashes, using pip-tools's pip-compile. In deployments, pip will now automatically check hashes of downloaded resources.
* Added requirements_txt_updater.sh which automated the creation of requirements.txt. Run this any time requirements.in changes.
* Added to our CI pipeline:
  - Check packages against pyup.io, which is a (one-month-delayed for free users) database of known Python package vulnerabilities.
  - Check packages against Pypi for newer versions. Added a new file requirements_txt_checker_ignoreupdates.txt that lists updated packages that we want to ignore in this check.
  - Check that requirements.txt is not out of date with requirements.in (i.e. someone edited .in but didn't regenerate .txt).
* Documented this in README.md.
* Check hashes of all dependencies fetched by fetch-vendor-resources.sh.

see #287

PR also removes `take_module_screenshots` command which hasn't been used for a while and had a dependency on a PDF-generating library not needed for anything else.


JoshData:
Greg asked me to write some additional thoughts about what I learned while putting this together.

Context
-------

This all began because we needed to get a handle on our software supply chain to address three concerns: a) What software is implicitly a part of GovReady-Q? b) Is Q resilient to upstream changes in that software (i.e. will Q break if software we rely on is changed or becomes unavailable)? c) Is that software safe to install in end-user internal networks (including client-side scripts via the browser)?

Q relies on many other software projects, from the operating system on up to client-side javascript. I began tackling this from the top of the stack long ago, moving remote/CDN-hosted code like jQuery and Bootstrap to be fetched at install time and statically served by Q at run time, instead of having browsers load them from CDNs (e.g. e3f07104addb1a68e6147f81a1e67b58be14953f, #209). That closed a run-time attack vector, e.g. what if the CDN began serving malicious content or simply wasn't available.

On the other end is the operating system, which we haven't yet addressed. The OS depends on the method of deployment. For instance, our production server runs RHEL (as provided by our ISP) but our Docker image is based on Debian (via the [Docker official Python image](https://hub.docker.com/_/python/)).

This pull request primarily addresses the middle of the stack: the software we bring in at install-time via pip/pypi (for Python packages) and via our fetch-vendor-resources.sh script (for other resources).

pip install
-----------

In a typical usage of `pip`, pip resolves and installs all dependencies of the packages that are requested to be installed, fetching the most recent version of a package that is in the range of permissible versions specified in package depedencies. This creates some problems in case the packages are modified --- packages we use during testing might not be the packages installed by end users if packages were updated upstream in the interim. New, backwards-incompatible or unsafe versions of packages could be brought in by pip's dependency resolving feature. Additionally, a vanilla usage of pip is susceptible to an attack where a package is replaced with malicious code (e.g. if the package maintainer's pypi credentials are stolen).

In 88e4d2baa2874df1c1b2b946ddba720109535666 I ensured all of our dependencies listed in requirements.txt were pinned to particular versions, but those packages didn't all pin *their* dependencies. In this PR, new tooling based on pip-tools's pip-compile script is added to flatten our entire set of dependencies into a static list where all packages are pinned to a particular version. We now store our top-level dependencies in requirements.in (where packages might not be pinned anymore). A new script calls pip-compile to perform pip's dependency resolution and store the complete list of pinned packages in requirements.txt, which we use both in testing and production to actually install the packages. Both files are stored in source code control.

pip-compile will also compute the SHA256 hash of each package it downloads an stores the hashes in reqirements.txt. pip version 8 and later will check that downloads match the hashes in requirements.txt if hashes are specified. If any hashes are found in requirements.txt, pip will require that every package it is installing be pinned and have a hash, so flattening our dependencies was a required first step before checking hashes could be added.

One gotcha here is that pip-compile will keep the pinned Python package versions that it sees when updating a requiremens.txt file that already exists. So although our requirements.in file no longer pins all of the packages (for convenience to make it easier in development to move forward to the latest version of packages), a vanilla call to pip-compile will not 'upgrade' the packages (i.e. write out newer version numbers and hashes). We specify --ugprade when we invoke pip-compile so that ignores the versions it last found and moves to the latest version of packages, to the extent the dependency isn't pinned to a version (or range of versions) by a package that depends on it.

safety
------

pyup.io has a Python package called safety that checks a list of Python packages for known vulnerabilties in their freemium/proprietary database. We now run the `safety` package during our build/test pipeline: during our Docker image build (in our Dockerfile), as a part of our continuous integration pipeline, and when requirements.txt is re-generated. The free version provides information from their one-month-delayed database.

We should consider paying for a pyup.io API key so that we get access to more timely information about vulnerabilities.

fetch-vendor-resources
----------------------

This PR also adds cryptographic validation of the resources we bring in via fetch-vendor-resources.sh. As suggested by #314, this PR now checks each downloaded file against a SHA256 hash hard-coded in the script. The script will abort with a non-zero exit code if a downloaded file does not match.

This script doesn't check if these upstream resources have newer versions --- that's something to come back to.

One concern here is in how we download fonts from the Google Fonts library. We download a third-party script at install time and validate its hash, and we use that script to download the fonts. Although the script does not know how to validate the font resources it downloads, we validate them after against additional hashes stored in the script. The fonts, however, are not versioned by Google, and it's not clear if Google ever updates the font resources in place. If the resources are updated, the hashes will no longer match and, without versioning, we will not be able to pick up an earlier version of the font to continue. Installations will block until the script is updated with new hashes.

JoshData:
This PR is now failing CI tests because in the last few hours a new version of Jinja2 was posted, and the failure signals that we should consider upgrading the package. So it already has helped!

PR #335, opened 2017-11-14
Jenkinsfile with Test stage

peterkaminski:
The Jenkinsfile with Test stage.  Other fiddly bits mixed in, unfortunately -- we can decide if we want to fix them in this merge or somewhere else.


peterkaminski:
thanks for the merge!

PR #342, opened 2017-11-21
Add GovReady logo to top nav

gregelin:
Small change to add GovReady logo to top nav bar.


JoshData:
Tests are failing because of upstream package changes, which I'll fix separately.

PR #350, opened 2017-11-29
 add a Dockerfile for CentOS 7

peterkaminski:
Because of a change in the `CMD` line for the virtualenv, I needed to make `dockerfile_exec.sh` executable (0775 vs 0644).  It probably should have been, anyway.

Tested and cleanly builds the image, and container runs GovReady-Q.  Only tested up to the GovReady-Q splash screen; @gregelin or @JoshData may want to run through more of GovReady-Q since it's running on a different OS.


gregelin:
I am not sure `dockerfile_exec.sh` should be `0755` because it means that anyone on the system could run the script and repoint the executing compliance app to a different data source or mount other volumes.

peterkaminski:
Cool, let's think about the permissions later.  I've reverted `dockerfile_exec.sh` to `0644` and now I call it via `source` instead.

PR #360, opened 2017-12-04
Ge erd2 (modifications to guidedmodules erd)

gregelin:
Modifications to Guidedmodules ERD to show relationships


JoshData:
You added django-database-storage-backend to requirements_txt_checker_ignoreupdates.txt but nowhere else, so that's probably a mistake. Can I remove that and merge?

PR #372, opened 2017-12-08
display PNG thumnails for HTML uploads on the question page, module-finished and review pages, and in output documents; fix PDF downloads of output documents

JoshData:


JoshData:
This is finally passing now.

PR #377, opened 2017-12-11
start apps should handle if an app fails to install

JoshData:
Better error handling when starting an app by "start apps."


gregelin:
This is nice new functionality. It would make a lot of sense to create a test for this.

PR #402, opened 2018-01-12
release tagging improvements, AppSource refactoring, impute condition error fix

JoshData:
* verify that release tags are PEP440 public version numbers
* verify that release tags are pushed to origin and aren't just local tags


JoshData:
@gregelin Need review.

PR #404, opened 2018-02-05
Various minor fixes

JoshData:
This PR has various small fixes including:

* updating some Python package dependencies
* fixing a CSS mistake
* fixing a Python indentation mistake
* adding error handling
* allowing the authoring tool to update a task in-place when a question is removed if the question has never been answered
* improving text rendering of multiple choice questions by using semicolons when the choices themselves have commas

### DIFF
--- a/guidedmodules/models.py
+++ b/guidedmodules/models.py
@@ -553,7 +553,12 @@ class Task(models.Model):
         # Check that all questions that need an answer have
         # an answer and that all module-type questions are
         # finished.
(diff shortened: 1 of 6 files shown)


JoshData:
Also discovered that django-allauth now requires Django 1.11 and we're still on 1.10, so I've added a line to requirements_txt_checker_ignoreupdates.txt so the build passes, and we should work on moving to the latest version of Django (1.11.10 or 2.0.2).

PR #410, opened 2018-02-07
Arrange projects by lifecycle stages on the projects page

JoshData:
This is a revamp of the layout of the Project Folders page, which now is the Projects page, to arrange Projects according to their status in a lifecycle such as the NIST RMF stages or e.g. the Software Development Lifecycle. Each Project's lifecylce stage is also shown on the main project pages under the project title.

The stage is determined by an output document named `govready_lifecycle_stage_code` in the project's root module. I made some fixes for using the template output type called "text" so that when we query this template we can run Jinja2 template commands but *not* run Markdown conversion to HTML (as we do with "html" output) since we're interested in a plain-text status code.

In addition, since folders are no longer shown on this page, the Project pages and the Task pages are revised to show context using standard Bootstrap breadcrumbs rather than the multi-line layout we had in which the Folder containing the Project was prominently shown.

The project icon's size was tweaked to better match the next next to it (which is either the project title or the project title + the module title if the strings differ).

And lastly this PR also contains the addition of a new button on the projects page labeled Add New System which goes to the Apps Catalog but filters on a new protocol ID minted for IT Systems.


siteapp/views.py
```diff
@@ -25,30 +25,90 @@ def homepage(request):
...
 
-    # Mark whether the user is an admin of any folders.
-    folders = list(folders)
-    for folder in folders:
-        folder.is_admin = (request.user in folder.get_admins())
+    # Load each project's lifecycle stage, which is computed by each project's
+    # root task's app's output document named govready_lifecycle_stage_code.
+    # That output document yields a string identifying a lifecycle stage.
+    for project in projects:
+        outputs = project.root_task.render_output_documents()
```
gregelin: Is calling `render_output_documents` going to force the top level app to render ALL output templates? Wouldn't that make determining the life cycle stage potentially very slow?

siteapp/views.py
```diff
@@ -25,30 +25,90 @@ def homepage(request):
...
 
-    # Count up the total number of projects in each folder. For admins, so they
-    # can know why they can't delete a folder.
-    from django.db.models import Count
-    folders = folders.annotate(project_count=Count('projects'))
+def assign_project_lifecycle_stage(projects):
+    # Define lifecycle stages.
+    # Because we alter this data structure in project_list,
+    # we need a new instance of it on every page load.
+    lifecycle_stages = [
```
gregelin: Seems like we are sticking configuration information inside the source code. Shouldn't we be able to define different life cycles for different users/orgs?

siteapp/views.py
```diff
@@ -25,30 +25,90 @@ def homepage(request):
...
+            project.lifecycle_stage = lifecycle_stage_code_mapping["none_none"]
 
-    # Sort the folders by name.
-    folders.sort(key = lambda folder : folder.title)
 
-    return render(request, "folder_list.html", {
-        "folders": folders,
+def project_list(request):
+    # Get all of the projects that the user can see *and* that are in a folder,
+    # which indicates it is top-level.
```
gregelin: Only top level apps in folders now?

siteapp/views.py
JoshData: The documents are actually lazy-rendered when they're accessed.

siteapp/views.py
JoshData: Yes.

siteapp/views.py
JoshData: Not sure what you're asking.

siteapp/views.py
gregelin: So we can treat this as temporary technical debt?

siteapp/views.py
JoshData: i.e. Only the one document that we need for the life cycle stage stuff will actually be rendered.

siteapp/views.py
JoshData: I think that's best. It would be easy to move this configuration somewhere else later, e.g. into the database and make it per-organization or something. I wouldn't call this tech debt - we're not making things harder later by leaving it here, we just haven't done the additional logic of getting this data from somewhere else yet.

PR #415, opened 2018-02-12
Polishes lifecycle stages css

gregelin:
Remove spacing between projects in lifecycle stages.


JoshData:
overflow: auto is usually hard for people to understand on mobile because there aren't scroll bars.

PR #422, opened 2018-02-14
New agile column layout for project pages

JoshData:
This PR makes some fixes:

* in how task progress is computed (typo in an earlier commit)
* in how binary assets are stored (Django 2.0 bug)
* removing dead code

And then the final commit:

* Re-vamps the project page changing the grid layout into an agile-style column layout.
* Updates the homepage lifecycle stages CSS styles to match ^.

### DIFF
--- a/templates/project.html
+++ b/templates/project.html
@@ -38,7 +38,8 @@
   padding-top: 10px; /* can't mess with side padding because it is a bootstrap column */
   padding-bottom: 10px; /* can't mess with side padding because it is a bootstrap column */
   margin-bottom: 15px;
-  border-radius: 5px;
+  border-radius: 3px;
+  border-bottom: 1px solid rgb(204, 204, 204);
 }
 
 .question-icon {
@@ -436,6 +437,7 @@ h3.question-group-title {
               </div> <!-- /question-group -->
             {% endfor %} {# group #}
 
+            {% comment %}
             {% if columns|length > 0 and column.groups|length == 0 %}
               <p style="line-height: 122%; font-size: italic; font-size: 95%;">
                 {% if column.has_tasks_on_left %}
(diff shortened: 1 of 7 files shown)


gregelin:
The text editor is not showing up for paragraph questions.

screen shot 2018-02-14 at 6 55 46 pm

templates/project.html
```diff
@@ -9,48 +9,56 @@
...
 
-.nav-tabs>li>a {
-  border: none;
-  color: rgb(125,125,125);
+.question {
+  background-color: white;
+  padding-top: 10px; /* can't mess with side padding because it is a bootstrap column */
+  padding-bottom: 10px; /* can't mess with side padding because it is a bootstrap column */
+  margin-bottom: 15px;
+  border-radius: 5px;
```
gregelin: Please modify border treatment:
```
border-radius: 5px;
border-bottom-style: 1px solid rgb(204, 204, 204);
```

gregelin:
It appears the cause for the failure to load quill.js and a number of other resources is the existence of the `local/environment.json` file in my install. When I remove the `environment.json` file (by renaming) and restart `runserver` all assets are loaded.

My guess is an issue related to the `host` key. Here is my `environment.json` file. I created it by copying the content recommended during the install process, then I set `debug` to false.

```
{
  "debug": false,
  "host": "localhost:8000",
  "https": false,
  "secret-key": "h8*(#roqomp8r41oq_js1=bm3fd+l6nuq$dovv36)cy4e8%(d4"
}
```

gregelin:
Content tweaks:

Remove text describing empty columns in stages and questions columns. Just leave blank. Delete:

```
% if column.has_tasks_on_left %}
                  There are no tasks in this column yet.
                {% else %}
                  There are no more tasks in this column. Congrats!
                {% endif %}
```  
in project.html

JoshData:
We diagnosed the problem above about the rich text editor not showing up as likely related to debug being set to False, which turns off the serving of static assets.

PR #429, opened 2018-02-18
Polish column and "card" UI styling

gregelin:
Make the column and "card" styling consistent across main pages.
Add box styling to columns for more visual delination.

Re-style context-bar to match focus-area background and
put context-bar on main pages.
Darken color of breadcrumbs text.

Move projects action buttons into context-bar.
Shrink page title on projects and add home icon.
Rename "Your Compliance Projects" title to "Assessments" as experiment.
Shrink width of module area displays on projects rows layout.
Add column styling to rows layout on project page for consistency.
Add back in "Finished" button, unable to display "In Progress" and "Start" buttons still.

Apply "card" styling to app-store.
Re-style app blocks in app-store to match "card" styling,
shift app-icon to right side of app block.
Remove vendor name from app block.
Re-style app-store top of page category links.
Move app-store search box to into context-bar.
Add home icon to app-store breadcrumbs.

### DIFF
--- a/templates/app-store.html
+++ b/templates/app-store.html
@@ -32,9 +32,12 @@ h2 {
     white-space: nowrap;
   }
 
-.app-category-nav a {
-  margin-left:20px;
-}
+  .app-category-nav *[data-category] {
+    margin-left: 20px;
+  }
+    .app-category-nav *[data-category].first {
+      margin-left: 0;
+    }
 
 .app {
   border-radius: 10px;
@@ -95,17 +98,19 @@ h2 {
 </style>
 {% endblock %}
 
-{% block contextbar %}
-    <div id="context-bar">
-    <div class="container-fluid">
+{% block contextbar %} 
+  {# kill breadcrumbs but add some space below site header #}
+  <div id="contextbar">
+    &nbsp;
+  </div>
+{% endblock %}
 
+{% block body %}
       <div>
-      {% block breadcrumbs %}
         <a href="/" style="color: black;">
           <span class="glyphicon glyphicon-home" style="margin-right: .25em"></span>
         </a>
         / Compliance Apps
-      {% endblock %}
 
       <div class="pull-right" style="margin-bottom: 1em">
         <form class="form-inline" onsubmit="return false;">
@@ -119,11 +124,6 @@ h2 {
         </form>
       </div>
 
-    </div><!-- container -->
-    </div><!-- div wrapper -->
-{% endblock %}
-
-{% block body %}
 
 {% if filter_description %}
 <p style="margin-bottom: 30px">These apps can help you with <i>{{filter_description}}</i>.</p>
--- a/templates/app-store.html
+++ b/templates/app-store.html
@@ -7,6 +7,12 @@ Compliance Apps
 
 {% block head %}
 <style>
+body {
+  padding-left: 20px;
+  padding-right: 20px;
+  background-color: rgb(247,247,247);
+}
+
 h1 {
   margin-bottom: 30px;
 }
@@ -25,19 +31,16 @@ h2 {
     color: #66C;
     white-space: nowrap;
   }
-    .app-category-nav > *[data-category]:before {
-      content: " · ";
-    }
-    .app-category-nav > *[data-category].first:before {
-      display: none;
-    }
+
+.app-category-nav a {
+  margin-left:20px;
+}
 
 .app {
-  border: 1px solid #D0D0D0;
-  /*border-top: 4px solid #88A;*/
-  box-shadow: 0 0 0 1px rgba(128, 128, 128, .1), 0 2px 3px rgba(128, 128, 128, .4);
+  border-radius: 10px;
+  border-bottom: 1px solid rgb(204, 204, 204);
   padding: 12px;
-  background-color: #FBFBFB;
+  background-color: white;
   margin-bottom: 24px;
   cursor: default;
   overflow: hidden;
@@ -55,7 +58,7 @@ h2 {
       }
     }
   .app .app-icon {
-    width: 30%;
+    width: 20%;
     max-width: 128px;
   }
     @media screen and (max-width: 768px) {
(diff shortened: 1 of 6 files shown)


templates/focus-area-styles.css
```diff
@@ -1,12 +1,17 @@
...
-  border-top: 4px solid rgb(74,164,74);
-  box-shadow: 0 0 0 1px rgba(0, 0, 0, .1), 0 2px 3px rgba(0, 0, 0, .2);
-  padding: 1.5em;
+  /*border: 1px solid #D0D0D0;*/
+  /*border-top: 4px solid rgb(74,164,74);*/
+  /*box-shadow: 0 0 0 1px rgba(0, 0, 0, .1), 0 2px 3px rgba(0, 0, 0, .2);*/
+  border-radius: 10px;
+  border-bottom: 1px solid rgb(204, 204, 204);
+  padding: 2em;
+  padding-bottom: 3em;
```
JoshData: IMO, dropping the green top border makes the page significantly less friendly.

templates/project.html
```diff
@@ -253,10 +268,12 @@ <h3 class="question-group-title">{{group.title}}</h3>
                     {# The layout of the project page is either a grid of components or a vertical list of components. #}
                     {% if layout_mode == "grid" %}col-sm-3 col-lg-2{% else %}col-xs-12{% endif %}
                     question
+                    question-row-container
```
JoshData: No need for this extra class, per above.

templates/project.html
```diff
@@ -50,18 +58,25 @@
...
 
 h3.question-group-title {
-  color: #777;
+  color: black;
+  font-weight: bold;
   padding: 16px 18px;
   margin: 0;
 }
 
+.question-row-container {
```
JoshData: This should just be on .question since `question-row-container` is on exactly the same elements as `question`.

templates/project.html
```diff
@@ -253,10 +268,12 @@ <h3 class="question-group-title">{{group.title}}</h3>
                     {# The layout of the project page is either a grid of components or a vertical list of components. #}
                     {% if layout_mode == "grid" %}col-sm-3 col-lg-2{% else %}col-xs-12{% endif %}
                     question
+                    question-row-container
 
                     {# Add CSS classes for whether this question is finished or if it's the next item for the user to do. #}
                     {% if q.is_finished %}task_finished{% endif %}
                     {% if q.first_start %}first_start{% endif %}"
+                    style="background-color: white;"
```
JoshData: This should be on the .question CSS class.

templates/project.html
```diff
@@ -353,14 +370,17 @@ <h3 class="question-group-title">{{group.title}}</h3>
                         {% if layout_mode != "grid" and not q.can_start_new_task %}
                           {% with task=q.tasks.0 %}
                             <div class="progress" style="width:80%; height: 5px;margin: 8px 0 0 0;">
-                              <div class="progress-bar progress-bar-success" role="progressbar" style="width: {{q.progress_percent}}%" aria-valuenow="{{q.progress_percent}}" aria-valuemin="0" aria-valuemax="100"></div>   
+                              <div class="progress-bar progress-bar-success" role="progressbar" style="width: {{q.progress_percent}}%" aria-valuenow="{{q.progress_percent}}" aria-valuemin="0" aria-valuemax="100"></div>
                             </div>
-                            {% if layout_mode == "row" %}
+                            {% if layout_mode == "rows" %}
```
JoshData: Ooops, good catch.

templates/project.html
```diff
@@ -353,14 +370,17 @@ <h3 class="question-group-title">{{group.title}}</h3>
                         {% if layout_mode != "grid" and not q.can_start_new_task %}
                           {% with task=q.tasks.0 %}
                             <div class="progress" style="width:80%; height: 5px;margin: 8px 0 0 0;">
-                              <div class="progress-bar progress-bar-success" role="progressbar" style="width: {{q.progress_percent}}%" aria-valuenow="{{q.progress_percent}}" aria-valuemin="0" aria-valuemax="100"></div>   
+                              <div class="progress-bar progress-bar-success" role="progressbar" style="width: {{q.progress_percent}}%" aria-valuenow="{{q.progress_percent}}" aria-valuemin="0" aria-valuemax="100"></div>
```
JoshData: This is meant to be deleted since you duplicated it in the if block below to only show progress if the question has been started but is not finished?

templates/projects.html
```diff
@@ -11,14 +11,14 @@
 
 {% block head %}
 <style>
-h1 {
-  margin: 15px 0 30px 0;
+body {
+  padding-left: 20px;
```
JoshData: Is this body side padding meant to be on all pages now? It should go into a single place like base.html.

templates/projects.html
```diff
@@ -11,14 +11,14 @@
 
 {% block head %}
 <style>
-h1 {
```
JoshData: See comment on h1 below.

templates/projects.html
```diff
@@ -60,11 +61,30 @@
       margin: 0;
       font-weight: bold;
     }
+
+  #context-bar {
+    background-color: white !important;
```
JoshData: You're adding a context-bar and then having to override its style so it doesn't look like a context-bar. Better to not create a context-bar in the first place (see below).

templates/projects.html
```diff
@@ -60,11 +61,30 @@
...
+  }
 </style>
 {% endblock %}
 
 <!-- Remove contextbar from top of page -->
-{% block contextbar %}{% endblock %}
+{% block contextbar %}
```
JoshData: Per above, better to keep this in body and not put it in contextbar but then not use contextbar's styles. I'm committing a change to put this back in body.

templates/projects.html
```diff
@@ -74,16 +94,6 @@ <h2>Complete your organization profile</h2>
...
 
-<div style="float: right; margin: 0 0 1em 5em;">
-  <a id="new-system" href="/store?protocol=govready.com/apps/compliance/2018/information-technology-system" class="btn btn-success"><i class="glyphicon glyphicon-plus"></i> Add system </a>
-  <a id="new-project" href="/store" class="btn btn-success"><i class="glyphicon glyphicon-plus"></i> Add other app </a>
-  {% if is_lonely_admin %}
-  <a href="#" onclick="show_invite_to_org(); return false;" class="btn btn-success"><i class="glyphicon glyphicon-send"></i> Invite colleague </a>
-  {% endif %}
-</div>
-
-<h1>Your Compliance Projects</h1>
```
JoshData: As above, every page ought to have an h1.

templates/focus-area-styles.css
gregelin: Josh, I agree the green border was nice and the revised color scheme is a bit gray. I've been experimenting with some further changes that fix these short comings. They are changes on top of this PR, so will make these first.

templates/app-store.html
```diff
@@ -92,35 +95,42 @@
...
+          </div>
+        </form>
+      </div>
+
+    </div><!-- container -->
+    </div><!-- div wrapper -->
 {% endblock %}
 
 {% block body %}
-<h1>
```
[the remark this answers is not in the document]
gregelin: Agreed. I removed the <h1> for styling. But you are right. We should put the <h1> back in and just style it appropriately for the page. I'll make an issue #430

templates/project.html
gregelin: My bad. Forgot to remove that css.

templates/project.html
gregelin: same.

templates/project.html
gregelin: Another artifact from an experiment.

templates/project.html
gregelin: Thanks.

templates/project.html
gregelin: Ok. I think I've cleaned this up.

templates/projects.html
gregelin: Moved.

templates/projects.html
gregelin: I think it is one half-dozen of the other. I was going for semantic consistency where the breadcrumb (and now action buttons) were always in the context bar and let the styling of the context bar change.

You're taking the path of displaying the same content (the breadcrumb) in the same location as the context bar bc it isn't styled the same. It's fine either way for the moment.

templates/projects.html
gregelin: Added `h1` back to `projects.html`

templates/projects.html
gregelin: Added h1 back into projects.html

templates/focus-area-styles.css
```diff
@@ -1,12 +1,14 @@
 body {
-    background-color: rgb(246,247,249);
+    background-color: rgb(247,247,247);
+    padding-left: 20px;
+    padding-right: 20px;
```
JoshData: Now that side padding is in body, it's redundant here.

PR #433, opened 2018-02-20
remove the external-function question type and calling Python functions defined in compliance app .py files from module templates

JoshData:
Long ago we created two features related to executing Python code written by compliance app authors:

* The `external-function` question type which runs a Python function when the question is "answered", instead of the user providing an answer.
* Calling Python functions from within module templates, like `{{ my_function(arg1, arg2) }}`.

These features both executed Python code defined in a .py file shipped along side a module YAML file, i.e. they are written by compliance app authors and not by GovReady.

Since running such Python code in the same process as Q requires that the code be fully trusted, both features have been disabled by default unless the `trust_assets` field was explicitly set to true on the AppSource that the compliance app containing the Python file came from.

It isn't practical for most end users to enable these features. This PR removes the two features.

### DIFF
--- a/docs/Apps.md
+++ b/docs/Apps.md
@@ -243,13 +243,11 @@ Apps can contain executable content (some of which is disabled by default):
 
 * JavaScript executed by the client browser served as a static asset and referenced by a `<script>` tag.
 
-* Python scripts executed by the GovReady-Q server for `external-function` questions.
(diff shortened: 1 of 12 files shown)


gregelin:
@JoshData Circle CI is having trouble downloading notifications for Django 2.0.

JoshData:
Yeah I will open a separate PR for that....

PR #441, opened 2018-02-22
add a new management command for taking screenshots of starting a compliance app and answering its questions

JoshData:


siteapp/management/commands/screenshots.py
```diff
@@ -0,0 +1,266 @@
...
+        os.makedirs(self.screenshot_basepath, exist_ok=True)
+
+        fn = os.path.join(
+            self.screenshot_basepath,
+            "{:03d}_{}.png".format(len(self.screenshot_image_filenames)+1, slug)
+            )
+        self.browser.browser.save_screenshot(fn)
+        self.screenshot_image_filenames.append(fn)
+
+    def login(self):
```
gregelin: Maybe we need to add explicit comments here about how we are addressing security? Or maybe this can only run when Debug is on?

PR #456, opened 2018-03-05
Improve project import results and add project metadata icon button

JoshData:
* After importing project data, the response is raw JSON. It was never finished. This puts it on an actual HTML page and makes the log messages more readable.
* An info icon is added to the projects listing page inside each project card to see some metadata about the project including the app version number, the project creation date, and the project root task last modified date.

### DIFF
--- a/guidedmodules/models.py
+++ b/guidedmodules/models.py
@@ -1044,9 +1044,9 @@ class Task(models.Model):
             # is a part of the same organization as the question it answers.
             task = Task.objects.filter(uuid=data["id"], project__organization=for_question.task.project.organization).first()
             if task:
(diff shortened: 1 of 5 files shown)


guidedmodules/models.py
```diff
@@ -1086,15 +1087,17 @@ def import_json_update(self, data, deserializer):
             # Only update the title if what's stored in the serialized data doesn't
             # match the automatically computed title.
             if data.get("title") and data["title"] != self.title:
+                deserializer.log("Updating title to {}.".format(data["title"]))
                 self.title_override = data["title"]
                 self.save(update_fields=["title_override"])
 
         # We don't chek that the module listed in the data matches the module
```
gregelin: Typo graphical error: "We don't chek"

PR #464, opened 2018-03-15
fixed related to using MySQL as a database

JoshData:
See individual commits.


gregelin:
Accepting pull request. The failed test is listed below and related to an "UNKNOWN" showing up on a license for a tool that we previously had listed as `CC0` license. See issue #465

PR #466, opened 2018-03-15
I broke docker when FIRST_RUN isn't set

JoshData:
In 56930725c9c72a5c25294b80ca9ad9bda5c272ca (#464) I added "set -euf -o pipefail" to the top of dockerfile_exec.sh, which is process zero inside the container, which tells bash to abort if any line fails. 

Besides that, it also checks for uses of shell variables that haven't previously been set (like undefined variables). To access environment variables in this mode when they may or may not be set outside of the script, it's necessary to access the variable with $(VARAIBLE-defaultvalue) syntax. In this case, use an empty default, i.e. $(VARIABLE-).


JoshData:
Tests passing now.

PR #471, opened 2018-03-19
go back to a plain text editor so autocompletes work

JoshData:
1) Reverts changes for the Quill text editor for discussions, going back to a simple textarea. (We're still using Quill for `longtext` questions, so we still have the Quill dependency.)

2) Expand the list of users shown in the @-mention autocomplete to include all users with read permission on Organization (i.e. anyone who can visit the organization site). When a user is @-mentioned who is not a discussion participant (currently defined as project members + guests), send them an invitation to join the discussion as a guest.

The second change seems to drastically slow down the loading of discussions, so I need to keep looking at that.

### DIFF
--- a/discussion/templates/discussion/discussion.html
+++ b/discussion/templates/discussion/discussion.html
@@ -4,10 +4,6 @@
 
 <script src="{% static "vendor/push.js" %}"> </script>
 
-<link href="{% static "vendor/quill/quill.snow.css" %}" rel="stylesheet">
-<script src="{% static "vendor/quill/quill.min.js" %}"></script>
-<script src="{% static "js/quill_commonmark.js" %}"></script>
-
 <style>
 #discussion-your-comment-tips {
   padding: 2px;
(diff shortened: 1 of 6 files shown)


discussion/templates/discussion/discussion.html
```diff
@@ -805,7 +827,7 @@
     method: "POST",
     data: {
       id: c.attr('data-id'),
-      text: open_editors[c.attr("data-id")].getContentsAsCommonMark()
+      text: c.find('.comment-edit textarea').val()
```
gregelin: Do we have a risk here of someone entering Javascript into the comment section and we just grab it? How do we make sure there is no dangerous Javascript in the comments?

JoshData: From a security standpoint this PR doesn't change anything since we're looking at client-side code.

My usual stance on unstructured user input is that it should be sanitized before it is displayed. Discussion comments are treated as CommonMark and are rendered using the CommonMark library with the "safe" flag turned on, which strips out raw HTML and unsafe link destinations.

gregelin: Thanks.

PR #473, opened 2018-03-22
speed up fetching @-mentionable users in discussions by batch loading

JoshData:
Getting the @-mentionable users in a Discussion (for autocomplete and then scanning the comment body for mentioned users) requires getting profile information from the users. Doing it user by user was slow even for a small number of users. This PR creates two major new methods for batch loading and has other small optimizations:
    
* Task.get_all_current_answer_records batch loads the current answers for one or more Tasks
* User.localize_users_to_org batch loads the account settings Tasks and their answers for one or more Users
* Task.get_or_create_subtask is optimized for when an answer already exists
* Discussion.attached_to is cached on the instance using a new wrapper property attached_to_obj

Also two small bug fixes.


gregelin:
Merging. Failed check is related to version of PIP.

PR #477, opened 2018-04-04
New settings page and various UI fixes

JoshData:


gregelin:
There is no way to set the display name in Django interface. Should we fix that in this PR?

gregelin:
There is no way to add another administrator to the organization in the settings page. Is that intentional?

PR #489, opened 2018-04-15
docker log path simplification, privacy policy and TOS pages, mysql documentation

### LINKED ISSUE #228: Add ToS and privacy policy
### Story
As a user, I want to understand GovReady's Terms of Service and Privacy Policy.

JoshData:
* Add stock privacy policy and terms of service pages, which can be overridden using the new branding app. Fixes #228.
* Change where our Docker deployment stores supervisord's logs to get around a permission denied error that I encountered on my desktop and in AWS ECS but not on my laptop.
* Note in the RHEL README that mysql-devel must be installed.

### DIFF
--- a/siteapp/views.py
+++ b/siteapp/views.py
@@ -1439,3 +1439,18 @@ def organization_settings_save(request):
         return JsonResponse({ "users": [user.render_context_dict(request.organization) for user in users] })
 
     return JsonResponse({ "status": "error", "message": str(request.POST) })
+
+def shared_static_pages(request, page):
+    from django.utils.module_loading import import_string
+    from django.contrib.humanize.templatetags.humanize import intcomma
+    password_hasher = import_string(settings.PASSWORD_HASHERS[0])()
+    password_hash_method = password_hasher.algorithm.upper().replace("_", " ") \
+        + " (" + intcomma(password_hasher.iterations) + " iterations)"
+
+    return render(request,
+        page + ".html", {
+        "base_template": "base-landing.html" if not hasattr(request, "organization")
+                         else "base.html",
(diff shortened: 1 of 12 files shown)


siteapp/views.py
```diff
@@ -1438,4 +1438,19 @@ def organization_settings_save(request):
...
\ No newline at end of file
+    return JsonResponse({ "status": "error", "message": str(request.POST) })
+
+def shared_static_pages(request, page):
+    from django.utils.module_loading import import_string
+    from django.contrib.humanize.templatetags.humanize import intcomma
+    password_hasher = import_string(settings.PASSWORD_HASHERS[0])()
+    password_hash_method = password_hasher.algorithm.upper().replace("_", " ") \
+        + " (" + intcomma(password_hasher.iterations) + " iterations)"
+
```
gregelin: Can you explain more about the password hasher?

templates/privacy.html
```diff
@@ -0,0 +1,22 @@
...
+		max-width: 70em;
+	}
+</style>
+{% endblock %}
+
+{% block body %}
+<div class="body-text">
+	<h1>Privacy policy</h1>
+	<p>The organization hosting this open source instance of GovReady-Q has not yet posted a privacy policy.</p>
+</div>
```
gregelin: Very nice touch.

siteapp/views.py
JoshData: The privacy policy we're adopting has a section that says what password hashing method is actually in use on the site. This fetches that information at run time rather than hard-coding it into the privacy policy document itself and sends the description of the hashing method to the template.

gregelin: Thanks

PR #501, opened 2018-04-29
fixes related to branding, docker, app README files, and deleting tasks

JoshData:
The first two commits update our Docker deployment to make it easier for downstream packagers to create images that extend our image, for instance an image that is customized to have organization-specific content.

I also added running "manage.py check" to the Dockerfile and to the Docker startup command so that we are checking for any Django-related problems during build and at startup. At startup we add `--deploy` which checks deployment-related settings and can only be run with actual deployment settings, so we can't run it at image build time.

The next two commits improve our deployment process in serverless environments by enabling HTTP->HTTPS redirects when the backend is handling both HTTP and HTTPS requests. Ideally the HTTP server takes care of this without HTTP requests getting to the Python process, but in some setups it's easier to just let Python handle that. The subsequent commit expands on our Docker on Amazon Web Services Elastic Container Service documentation.

The next three commits are improvements for app authoring. The first begins reading a new README.md file in compliance apps and uses this for the long description in the app catalog. The second commit allows Jinja2 tags to be multi-line --- Jinja2 allows this but our templating system didn't always allow it because multi-line Jinja2 tags were not shielded (so to speak) from the Markdown rendering which occurs first. The third commit improves error messages for template Jinja2 syntax errors by adding a line number (which is relative to the start of the template string, not to the top of the YAML file, unfortunately).

The next commit improves the API POST method (both the POST-as-form and POST-with-JSON) to allow updating the answers of `module-set`-type questions, which wasn't previously implemented. We create new Task answers when using the API to update these questions.

Finally, because the API now creates new Tasks on each POST that references a `module-set` question, we needed a way to delete these Tasks from the Project. So I (re-)implemented a task deletion button, which doesn't actually delete any data from the database but marks the Task as deleted and un-links it from questions it is an answer to by saving new answers to those questions with the same value but the task removed. We should really be doing something similar for Project deletion because right now we're doing a cascaded delete of the Project and its root task, which is destructive and will silently change question answers that referred to the project/root task contrary to our goal of having Task answer histories be immutable.


guidedmodules/views.py
```diff
@@ -1023,22 +1024,43 @@ def authoring_edit_module(request, task):
...
+    # Project admins can delete tasks. It's not a destructive
+    # operation at the database level, but it does cause the Task
+    # to become inaccessible.
+
     if request.method != "POST":
         return HttpResponseNotAllowed(["POST"])
 
     task = get_object_or_404(Task, id=request.POST["id"], project__organization=request.organization)
-    if not task.has_write_priv(request.user, allow_access_to_deleted=True):
+    if not request.user in task.project.get_admins():
```
gregelin: It seems that has_write_priv is mor granular than get_admins. Can you elaborate?

gregelin: Thanks.

PR #524, opened 2018-06-13
Vagrantfile

JoshData:
git clone https://github.com/govready/govready-q
	cd govready-q
	vagrant up
	vagrant ssh -- -t manage first_run

Then visit http://192.168.150.150:3031/ in your browser!


gregelin:
Checks were fixed in #525

PR #531, opened 2018-07-06
refactor module_sources.py splitting it into two new files

JoshData:
The module was getting big. This splits it into two smaller modules, one that provides abstract classes for getting compliance apps from remote sources and one that loads the apps into the database.

This PR has commits from the other pending PR. I'll fix that when the other PR is merged.

### DIFF
--- a/guidedmodules/app_source_connections.py
+++ b/guidedmodules/app_source_connections.py
@@ -1,25 +1,46 @@
-from django.db import transaction
-from django.db.models.deletion import ProtectedError
+###########################################################
+# App Source Connections
+#
+# This module defines the abstract AppSourceConnection
+# class and its subclasses which fetch app definitions from
+# local directories and remote locations like git
+# repositories.
+#
+# The main implementation is the PyFsAppSourceConnection,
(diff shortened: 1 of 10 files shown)


gregelin:
Wow, lots of work in this PR. I'd love to be walked through it. Does modification also update documentation (I haven't read yet).

JoshData:
Most of the commits are what you merged in the last PR. I've force-pushed this PR to kill those now that they're merged. Tests will run again...

guidedmodules/app_source_connections.py
```diff
@@ -0,0 +1,611 @@
...
+
+
+class GitRepositoryFilesystem(SimplifiedReadonlyFilesystem):
+    def __init__(self, url, branch, path, ssh_key=None):
+        self.url = url
+        self.branch = branch or None
+        self.path = (path or "") + "/"
+        self.ssh_key = ssh_key
+
+        # Create a local git working directory.
```
gregelin: Maybe if we added here " # Create a local git working directory that will be automatically destroyed after use."

PR #544, opened 2018-08-06
store output documents in separate files from the YAML files where they are referenced

JoshData:


docs/source/Schema.md
```diff
@@ -123,15 +123,31 @@ Projects can provide sets of exemplar answers for use in test scripts. e.g.:
 Documents
 ---------
 
-Documents occur as `introduction` and `output` documents of Modules as well as Question prompts. A document appears as:
+Documents occur as `introduction` and `output` documents of Modules, and a restricted for of documents also occurs in Question prompts. A document appearing in the output documents list is given as:
```
gregelin: Typo in sentence? "and a restricted for of documents" doesn't make sense.

JoshData: Should be "form." Fixing...

PR #557, opened 2018-10-23
Update change log

gregelin:


JoshData:
We never released 0.8.3 so we should just change the date on 0.8.3 to today, and move "Add system minimum and recommended software requirements to documentation." to the 0.8.3 list. (The other two are already covered by the 0.8.3. notes.)

gregelin:
In that case, we need to skip version 0.8.3 and go to 0.8.4 because I put in my monthly report to DHS S&T that we  were releasing v08.8.4. I'll update the CHANGELOG and update tag to -rc1.

JoshData:
Roger that. I will merge and try a new build of the docker image in a moment....

JoshData:
I removed the last two bullets because they were redundant....

gregelin:
OK. Are we ready to merge?

JoshData:
Yes, I'll merge now. I'm going to test the build, and after that I can publish the new image to Docker Hub and update our production instance.

PR #559, opened 2018-10-24
add VERSION file and build checks

JoshData:
Our Docker Hub nightly builds broke because our Dockerfile expects a `VERSION` file which is not in our repository but has been created with our build script `docker_image_build.sh` during out release process.

This PR puts the `VERSION` file into the repository and adds some checks to ensure we keep it up to date:

* The version in the file must be a valid [PEP 440 public version identifier](https://www.python.org/dev/peps/pep-0440/#public-version-identifiers). This is checked on each CircleCI build.
* The version in the file must exactly match the version number of the latest (i.e. topmost) release mentioned in `CHANGELOG.md`. This is checked on each CircleCI build.
* If `CHANGELOG.md` has a section at the top called "In Development," then the version in the `VERSION` file must have `+devel` appended to it so that the version doesn't continue to use the last release's version number in subsequent pre-release builds. (The `+` signals a [PEP 440 local version identifier](https://www.python.org/dev/peps/pep-0440/#local-version-identifiers).) This is checked on each CircleCI build.
* In releases, the version number in the `VERSION` file must match the git tag of the release. This is checked when making a new release with `docker_image_build.sh`.

This PR may need some more testing before we merge. We need to check that everything still works fine:

* in a release (the `VERSION` file has a second line with the git commit hash in this case)
* in a local checkout of the source code (the application uses git at run time to fetch the commit hash)
* in a docker build (git commit hash is not available)


gregelin:
@JoshData Thanks for doing this!

gregelin:
@JoshData how do we "This PR may need some more testing before we merge"? Is it just running a build for Docker?

gregelin:
@JoshData I see a successful test on jt_primary branch in https://hub.docker.com/r/govready/govready-q-nightly/builds/. I think that is sufficient test.

screen shot 2018-10-24 at 7 52 59 am

JoshData:
I just tried the nightly build and it seems fine, so I guess we're good.

PR #560, opened 2018-10-30
First pass at deployment with Terraform

peterkaminski:


gregelin:
@peterkaminski Where do you think we are with this merge request? Is it still relevant?

peterkaminski:
@gregelin Unless we have some other Terraform work that would supersede it, yes, I think it's worth keeping.  Probably it should be checked to make sure it's still working, then merged.

gregelin:
Temporarily closing this PR until 0.9.0 is released.

Issue #563, opened 2019-01-17
Examine Django security issues CVE-2019-3498 More information


gregelin:
# Remediation

Upgrade django to version 2.0.10 or later. For example:

```
django>=2.0.10
```

# Details

[CVE-2019-3498](https://nvd.nist.gov/vuln/detail/CVE-2019-3498) (low severity)

Vulnerable versions: >= 2.0.0, < 2.0.10
Patched version: 2.0.10

In Django 1.11.x before 1.11.18, 2.0.x before 2.0.10, and 2.1.x before 2.1.5, an Improper Neutralization of Special Elements in Output Used by a Downstream Component issue exists in django.views.defaults.page_not_found(), leading to content spoofing (in a 404 error page) if a user fails to recognize that a crafted URL has malicious content.

gregelin:
Fixed back on Jan 19, 2019 with PR #564

PR #564, opened 2019-01-17
update to commonmark 0.8.1

JoshData:
@gregelin experienced https://github.com/rtfd/CommonMark-py/issues/134 in commonmark 0.8.0.

Our package https://github.com/GovReady/CommonMark-py-Extensions and my package https://github.com/if-then-fund/django-html-emailer both needed updates to be compatible with commonmark 0.8.1. I updated them and then bumped our dependencies by running our `./requirements_txt_updater.sh` script.


JoshData:
This also fixes #563 because I updated all of the dependencies.

JoshData:
The build failures (unrelated to CommonMark) are now fixed by pinning one of our dependencies to an earlier version (see https://github.com/pallets/markupsafe/issues/108) and moving from Python 3.4 to Python 3.6 on CircleCI, to match the version of Python I use in development when preparing requirements.txt. I also updated other dependencies, removed unnecessary code that has moved into one of our dependencies, and fixed a security issue in rendering potentially unsafe CommonMark.

Issue #581, opened 2019-04-09
Thorough review and update existing deployment instructions


gregelin:
We have documentation for various installation and deployment. Documentation s old and needs to be reviewed, updated or retired as appropriate.

---
Components to this task (edit by Ethan):

- [x] Split off OS-specific and application-specific instructions
- [x] Write doc page for prod vs. dev
- [ ] Pull out needed details from Docker page
  - [ ] mention automated tests on the dev deployment page
  - [x] figure out this mailgun thing? or create a separate issue for it
- [ ] Check and update prose to account for reorganization (e.g., explain why Postgres section is `yum`-heavy, OR make it more neutral on distro)

ethankaminski:
Three things I needed to install, which weren't mentioned, were the `Python.h` file (should be mentioned), `gcc` (should be mentioned, although I expected this would be needed), and `mysql_config` (covered by issue #505). That's something to include as part of the documentation update.

ethankaminski:
Oh right, and other things I've got in my notes:

* It would be good to provide `pip3 install --user` as the snippet, because this seems to be recommended (by a first-party warning message) as better than `sudo`
* There are at least 3 sets of Linux instructions, so it would be good to check those for unintended discrepancies

ethankaminski:
Note from a call: part of this will be explaining weird "gotchas" and such (one oddity that pops out to me is why `xvfb` is required - it's a Debian workaround).

gregelin:
@ethankaminski @peterkaminski I am wondering if we should begin to split out instructions by OS?

ethankaminski:
@gregelin Yeah, splitting the instructions in a more logical/deliberate fashion is something that struck me as a good idea, after I combed through the docs yesterday to get a sense for what the current state is.

Another thing that might seems worthwhile is organizing the docs a bit more by task, and also bringing the OS pages closer into sync. Currently, I noticed that each OS's page is opinionated about stuff like which public-facing webserver is used (_differently_ opinionated, even), which seems a bit odd.

An approach I was considering is spliting off things like DB-specific configuration or webserver-specific  configuration(Apache/Nginx) into their own pages, where possible, and then referencing those from the appropriate OS page.

Also, FWIW, I won't be able to test Mac or Windows instructions, but I can do most flavours of Linux pretty readily.

gregelin:
@ethankaminski That all makes sense. Maybe propose an outline within this thread, or in a linked wiki page and we can discuss.

[...]

PR #593, opened 2019-04-17
Split requirements file based on optional modules (fixes #505)

### LINKED ISSUE #505: Put database requirements into separate requirements_db.txt files
Users should not have to install unnecessary database libraries on their laptops in order to use GovReady-Q. We should create separate requirements_mysql.txt, requirements_psql.txt files for the relevant libraries.

ethankaminski:
This addresses the DB-related modules issue, from #505.

Note: I made a few judgement calls when choosing where to add a mention of "install this other file too" vs. just removing the `mysql` reference. When a holistic doc review is done (issue #581, which it sounds like I'll end up doing), that's perhaps something to pay attention to.

Note2: I did not update anything in `deployment/pws/`, since I couldn't quickly tell whether that needs MySQL or not.


gregelin:
@ethankaminski Very glad you found this issue. The poor indexing on the levels bothered me, too. Was not sure how long the hunt would take to resolve.

ethankaminski:
@gregelin The `<h#>` issue in the other PR, you mean? Yeah, on my second pass through the docs, that stuck out rather obviously - at first I thought I needed to add section headers, until I realized that there was a conversion error. Ended up being something like 30m - 60m to hunt down (I think), once I noticed the issue.

PR #597, opened 2019-04-20
Misc merge commits with no file changes.

gregelin:
Cleaning up various commits against master.


gregelin:
@peterkaminski Also opened to alternative remedies for these merge commits.

peterkaminski:
Looks like the reason people don't like merge-only commits is because it makes the logs less clean:

https://softwareengineering.stackexchange.com/a/218807

The way to avoid them is to use `rebase`:

http://kernowsoul.com/blog/2012/06/20/4-ways-to-avoid-merge-commits-in-git/

If you need to rebase for a commit that's not the head, try this:

https://stackoverflow.com/q/7744049

Since it's not a big deal, let's just merge this PR, and look into using `rebase` more/better in the future.

PR #600, opened 2019-04-21
Refactor documentation...

gregelin:
Reorder main index to include Deployment documentation second.
Move enterprise single-sign-on into own document.
Move system requirements to own document and organize
under installing and deploying. Refactor headings to
use "#" indicators instead of underlines.


gregelin:
@peterkaminski Just documentation changes here.

PR #608, opened 2019-04-25
Production and Development environment docs

ethankaminski:
This splits off some documentation into pages specific for Production environments and Development environments, as part of #581. It also creates a "generic Unix/POSIX" page, because that information was part of the "local dev" deployment instructions.

### DIFF
--- a/docs/source/Deploy.md
+++ b/docs/source/Deploy.md
@@ -9,6 +9,7 @@ Deploying GovReady-Q
    deploy_docker.md
    deploy_host_os.md
    deploy_prod.md
+   deploy_local_dev.md
    configure_db.md
    configure_webserver.md
    Environment.md
--- a/docs/source/Deploy.md
+++ b/docs/source/Deploy.md
@@ -8,6 +8,7 @@ Deploying GovReady-Q
    requirements.md
    deploy_docker.md
    deploy_host_os.md
+   deploy_prod.md
    configure_db.md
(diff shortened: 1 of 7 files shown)


gregelin:
@ethankaminski So I'm the one who suggested that we provide instructions for different environments. As I'm reading the changes, I'm realizing that approach necessitates having the command-line instructions repeated in multiple places creating a potential violation of Don't Repeat Yourself principle.

I'm also realizing that we are making enough changes to the content  that we need to version our docs.

I do like a the language and description improvements.

In short, I need to build the docs and look at a few examples online of instructions to decide how to best balance the competing value of DRY and specificity per context.

ethankaminski:
@gregelin Sphinx appears to provide a way to embed (not just link) one document into another, if your concern about DRY is related to the same text being written in multiple places in the source. That would probably be a very good idea for the "next steps" section I've created, for instance, and I could use that as a way to test the functionality.

If you're concerned about having slightly-different versions of the same thing written in different places, I see a small number of things that are obvious candidates for removal. However, if this is your concern, it would also require some additional thought, as there are plenty of similar pieces of information that don't look all that easy to remove. Although, maybe some of those could be made identical rather than similar, and then handled by embedding.

gregelin:
@ethankaminski Knowing we can include with Sphinx is helpful. Accepted the PR and we can continue to refactor the docs and improve.

PR #611, opened 2019-04-26
Improve skip button feature and show answer metadata in document templates

JoshData:


guidedmodules/module_logic.py
```diff
@@ -1409,6 +1420,62 @@ def choices_not_selected(self):
...
+
+    @property
+    def skipped_reason(self):
+        if self.answerobj is None:
+            return self.answerobj
+        return self.answerobj.skipped_reason
+
+    @property
+    def unsure(self):
+        # If the question was answered by a user, return its unsure flag.
```
gregelin: These need to be edited

gregelin: I edited them.

Issue #607, opened 2019-04-25
Document all environment.json keys, centrally


ethankaminski:
The documentation for `environment.json` keys is scattered around various pages (e.g., I only just found `mailgun_api_key` today, in the RHEL/Centos page). A list of all the keys, and what they do, is the sort of thing that should go in the current Environment Settings page.

ethankaminski:
Some handy `grep` commands for this (so I can remember them):

```
# get each top-level key, by itself, no dupes
grep -rP -ho "environment(\.get\(|\[)(['\"])\K.*?(?=\2)" siteapp/ | sort | uniq
```
```
# show key in original context
grep -rnP -o "environment(\.get\(|\[)(['\"])\K.*?(?=\2)" siteapp/
```

gregelin:
This has been completed.

Issue #589, opened 2019-04-12
Reassess default option of localhost:80


ethankaminski:
If the org subdomain configuration option is omitted, Q will automatically pick up `localhost:80` as the default option. This is probably a bad choice, because port 80 is privileged (requires root), may already have something running on it, and the use of "localhost" means that the user can't easily run a reverse proxy in front.

Using `localhost:8000` might work, because that seems to be the port that Q binds to by default; or perhaps error out, requiring that an option be set.

(note: resolving #584 might obviate the need for this issue)

ethankaminski:
Addressed by #612.

PR #612, opened 2019-04-27
Change default value for org subomains to match `host` setting

ethankaminski:
This should address the main aspect of #589 (namely, the fact that the default value is probably never useful). I know that's marked as "best" priority, but this commit was trivial to write while working on #607 and #581.


gregelin:
The above commit caused a problem in the `localhost:8000` case by failing to strip off the port.  GovReady-Q would throw a *DisallowedHost* error for local installs where GovReady-Q was generating default values (to help guarantee a successful first run).

Correcting issue. And also adding documentation regarding the *DisallowedHost* error because it is a common error. Also fixing a comment typo.

PR #618, opened 2019-04-29
Fix documentation problems in v0.8.6

gregelin:
Fix some repeated copy in v0.8.6 and also adjust the tone
in the explanation of the skip button feature because it
sounded snarkier than intended.


gregelin:
I prefer the version "...assessments that cannot be changed easily and it is better to have users get stuck..." than the version "assessments that cannot be changed easily, when it is better to have users get stuck".  The "when" implies assessments that cannot be easily changed creates the situation of it being better to have users get stuck.

gregelin:
Making the change...

PR #621, opened 2019-05-01
Document `local/environment.json` keys

ethankaminski:
This creates a list of all top-level keys used by `local/environment.json`, along with descriptions for some of them.

I considered a table instead of a bulleted list, but the current CSS doesn't seem to work well with descriptions in tables (long prose doesn't auto-wrap). Plus, a list will be more-easily expanded to nested keys, like the subkeys of `email`.

I could also add descriptions for all keys (instead of just some), if desired. But, I'll open this as-is, because it still seems useful.

### DIFF
--- a/docs/source/Environment.md
+++ b/docs/source/Environment.md
@@ -1,5 +1,25 @@
 ## Environment Settings
 
+### top-level keys
+
+- `admins`
+- `branding`: used for [custom branding](Custombranding.html)
+- `db`: if supplied, this is the DB connection used. See [DB configuration](configure_db.html).
+- `debug`: should be `false` or absent in production environments. If set to `true`, turns on certain debug/development settings.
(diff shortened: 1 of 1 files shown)


gregelin:
@ethankaminski I added information to variables that did not have explanations. Then I noticed @JoshData had also documented many of these variables on the page https://govready-q.readthedocs.io/en/latest/deploy_docker.html#environment-variables-for-launching-the-container-without-our-run-script.  

It's probably OK to  have explanations in both places, but we should synchronize the content to the extent it makes sense to do so and cross link the two sections to each other.

ethankaminski:
Oh, good point, those _would_ have to be configured using that JSON config file. I saw "environment variables" (i.e., ends up accesible in `ENV`) rather than "config values", and at the time filed it into a different mental folder.

That might be another case where it's worth using Sphinx includes, then.

Issue #590, opened 2019-04-12
Make environment.json creation a little less perplexing during installation


ethankaminski:
During installation, it feels odd to receive an informational message with suggested `environment.json` settings at the _top_ of a big block of informational messages.

One idea that could feel less odd is to prompt the user if they want to populate the file with some useful default settings, rather than emitting that message and then carrying on with no pause.

gregelin:
Closed due to age.

PR #622, opened 2019-05-06
Prompt for auto-creation of environment.json when run in a TTY

ethankaminski:
A prompt of this sort is useful for getting a user's first dev/trial environment set up quickly, and for cases where it's more convenient to not have to copy-paste. (e.g., my workflow often involves a lot more keyboard usage than mouse usage, for instance)

Addresses #590, though I can think of other ways to address that, if this approach isn't preferred.

### DIFF
--- a/siteapp/settings.py
+++ b/siteapp/settings.py
@@ -46,6 +46,23 @@ else:
 	print(json.dumps(environment, sort_keys=True, indent=2))
 	print()
 
+	# If we're in an interactive session, prompt the user if they'd like to write out these defaults
+	from sys import stdin
+	if stdin.isatty():
+		# should be safe to omit any mention of prod vs. dev environments, because that
+		# `localhost:8000` would need to be edited anyhow
+		print("Would you like to create this file automatically, with the above content? (Y/N)")
+		answer = input("> ")
+		if answer[0].upper() == 'Y':
+			print("Creating `local/environment.json` file with default values")
+			fh = open(local("environment.json"), 'w')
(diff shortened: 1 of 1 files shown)


gregelin:
Hi I am remembering more details about the creation of this file. We need to not create this file by default so we can make a read only File system Docker version. We need to not prompt user so an automated installed will not hang waiting for input. I wonder if we could create a command line flag to create the environment as part of an instr I wonder if we could create optional command line flag to create the environment.json as part of an  install?

ethankaminski:
I considered default behaviour and automated installs when coding this, and accounted for that by defaulting to "no" (so, matching current behaviour), and also by prompting for input only when run in a TTY (i.e., interactive session). To the best of my knowledge, automated installs in Docker should return `False` when calling `stdin.isatty()`.

However, it looks like CircleCI doesn't like the use of `input` due to a Python2 security risk, so I'll get that changed tomorrow to use a different call.

gregelin:
@ethankaminski I think we are not going to auto-create the `environment.json` file. I think we will in fact eventually remove the reminder message once our documentation improves. So we won't do this PR and you do not find a different solution.

ethankaminski:
Makes sense; that approach would also result in there no longer being a reminder-prompt in an awkward spot during initial setup. I'd say it's useful to retain the one-line reminder that environment.json is missing; it's more the "here are some defaults to copy-paste" part that struck me as timed awkwardly.

gregelin:
Agreed!

Greg Elin
CEO, GovReady PBC
p: 917-304-3488
e: gregelin@govready.com

PR #625, opened 2019-05-09
Include MySQL required libraries in Docker deployment

gregelin:
We broke out the mysql requirements so it would not be necessary
to install mysql if one was just using sqlite in #593, but
then did not update `Dockerfile` to pull in separated
MySQL requirements. This commit fixes that error.

Likely error for users of Docker container trying to connect to MySQL:

```
Starting at XXX.XXX.XXXX:443 with HTTPS true.
Traceback (most recent call last):
File "/usr/lib64/python3.6/site-packages/django/db/backends/mysql/base.py", line 15, in <module>
import MySQLdb as Database
ModuleNotFoundError: No module named 'MySQLdb'
The above exception was the direct cause of the following exception:
Traceback (most recent call last):
File "manage.py", line 10, in <module>

...snip...

r/lib64/python3.6/importlib/__init__.py", line 126, in import_module
return _bootstrap._gcd_import(name[level:], package, level)
File "/usr/lib64/python3.6/site-packages/django/db/backends/mysql/base.py", line 20, in <module>
) from err
django.core.exceptions.ImproperlyConfigured: Error loading MySQLdb module.
Did you install mysqlclient?
```

Error caused by our splitting the mysql-related requirements into a separate
 `requirements_mysql.txt` file in #593 and then not updating our Dockerfile to pip install the mysql requirements.

We broke out the mysql requirements so it would not be necessary to install mysql if one was just using sqlite. See [https://govready-q.readthedocs.io/en/latest/deploy_rhel7_centos7.html?highlight=mysql#preparing-q-source-code](https://govready-q.readthedocs.io/en/latest/deploy_rhel7_centos7.html?highlight=mysql#preparing-q-source-code)

```
# if you intend to use optional configurations, such as the MySQL adapter, you
# may need to run additional `pip3 install` commands, such as:
# pip3 install --user -r requirements_mysql.txt
```

Recommended fix is to build the Docker image with MySQL by changing lines 28 - 33 of `Dockerfile` have been updated to read:
```
# Copy in the Python module requirements and install them.
# Manually install database drivers which aren't in our requirements
# file because they're not commonly used in development.
COPY requirements.txt ./
RUN pip3.6 install --no-cache-dir -r requirements.txt
RUN pip3.6 install --no-cache-dir -r requirements_mysql.txt
```


gregelin:
@ethankaminski Can you please build the docker container from the Dockerfile in this PR and test GovReady-Q works using a MySQL database?

ethankaminski:
Can do, I'll be able to sit down with that in the next few hours.

ethankaminski:
@gregelin It needed two more lines, which I've added in a commit to this branch. Had to copy over the new .txt file to the Docker filesystem, and had to re-add the line where it installs the `mysql-devel` package.

gregelin:
Thanks.

Greg Elin
CEO, GovReady PBC
p: 917-304-3488
e: gregelin@govready.com

PR #637, opened 2019-06-14
Ge docs 0.9.0 announce2

gregelin:
Include adding App Sources instructions to 0.9.0 page. Having the complete instructions in one place will make it easier to install and configure 0.9.0.


gregelin:
Because this merge is just documentation and is needed for a deadline, I self-merged.

PR #650, opened 2019-06-19
Test Data Population commands

ethankaminski:
Adds a few new `manage.py` commands to automatically populate data into a Govready-Q 0.8.6 instance, stored in a new top-level directory (`testmocking`).

* `populate`: creates test users and test organizations
* `clear_mock_data`: wipes data created by `populate`. not yet tested in conjunction with `answer_task`.
* `answer_task`: automatically fills in questions in a form section (look for "`/tasks/[id]/" in the URL bar, then grab that ID)
* `inspect_task`: debug-ish info on a specific task. Mainly for me to sanity-check things, not needed if you want to stand up a new test org.

In a future cleanup pass, one thing that might be worth changing is how the `populate` and `clear_mock_data` store IDs of the generated data. Currently, those IDs are going into files prefixed with "TMP_" in the new `testmocking` directory; would be better to put it into a proper data directory, or a proper temp directory, or a separate DB table, or something. However, these files are still segregated from the actual app source code, so it seems safe to leave them there while a decision is made on where they should actually live (or whether `clear_mock_data` is needed at all - though I, personally, do find it useful).


gregelin:
@ethankaminski Docker Hub failed building the image.

```
ModuleNotFoundError: No module named 'testmocking'
```

I think we need to include libraries into `requirements.in` and `requirements.txt`?

See: https://cloud.docker.com/u/govready/repository/docker/govready/govready-q-nightly/builds/b92bce1b-488f-4f6a-a56f-1ce3198d7140

ethankaminski:
@gregelin I see what's likely causing that. I think we actually need to add it to the Dockerfile. This new module will be similar to `guidedmodules`, which doesn't show in either of the `pip` requirements files, but does show in `Dockerfile`.

I've added a new `COPY` command to that Dockerfile, which looks like it will resolve the issue with Docker Hub not building successfully. If it doesn't, then I'll troubleshoot further.

Incidentally, it looks like some CircleCI config is also based on whitelisted modules. Test coverage may not be necessary in this directory, but would it be worth adding static analysis on this code? (either now, or in the next round of work on this feature)

ethankaminski:
@gregelin heads up that I've added `parsel` (fairly reputable, it looks like) to the `requirements_in.txt` file. This lets me manipulate HTML (without JS), without needing the full Selenium overhead.

I also ended up going with a web-based approach for starting new assessments. It looks like the `module` question type takes a few more steps than just "create 1 DB entry -> done", so in the meantime, I've added this web-based version just to get something functional in place.

My plan from here is to (1) tie everything together into an end-to-end "just go" function; (2) clean up some of this code (e.g., names instead of random words, or clean up the HTML logic, etc.); (3) if time permits, get a non-HTTP version written, because one of these pages is very slow to execute. Good chance of getting (1) and (2) done over the weekend, unsure about (3).

gregelin:
@ethankaminski I would not currently worry about item (3) the non-html version.
I will do merge after "just go" and human names clean up complete.

ethankaminski:
@gregelin Got those done. The command `python3 manage.py populate --full` generates new orgs (default 1 org) with assessments filled out. Takes about 2 minutes on my system.

I pre-generated a list of names from generatedata.com and dropped that in as a data file, for the usernames. For organization names, I observed that generatedata.com spits out lorem ipsum, so IMO it will be more readable to just stick with dictionary words (though I did follow their lead and added a company-ish suffix on the end, like "Ltd").

gregelin:
There's one TODO in the file `testmocking/web.py`":

```
def add_system(self):
  # TODO see if this shouldn't be hardcoded
  self.load("/store?protocol=govready.com/apps/compliance/2018/information-technology-system")
```

We could add multiple protocols that work.

Issue #643, opened 2019-06-18
Merge pk-nginx-20190605 into 0.9.0.dev


peterkaminski:
The Q+nginx branch `pk-nginx-20190605` is done, including docs.  Merge the `pk-nginx-20190605` branch into `0.9.0.dev` to add it to the 0.9.0 release.

gregelin:
@ethankaminski Can you please test these instructions to make sure they work?

https://github.com/GovReady/govready-q/tree/pk-nginx-20190605/deployment/docker-compose-nginx

ethankaminski:
Can do, probably in several hours.

gregelin:
Merged.

PR #651, opened 2019-06-20
Pk nginx 20190605

gregelin:
Merge in documentation for using docker-machine to deploy GovReady-Q with NGNX.


gregelin:
This PR will address issue #643

ethankaminski:
Doing a quick sanity check on whether there's much risk in including a statically generated private key, let's see:

Since this is a cert-signed cert and _not_ a root cert, (a) any reasonable user agent would freak out if you told it to fully trust the cert (as opposed to just ignoring the error temporarily), and (b) there's likely not any way to use this to issue certs on other domains, in the first place.

Cool, I think that aspect passes a sanity check (it still gives me pause, but in a code smell sense rather than an actual security risk).

PR #655, opened 2019-06-28
Link to existing App Source docs rather than repeat, to emphasize changes

ethankaminski:
This replaces a large block of repeated content in the 0.9.0 docs with a link to the existing version. Specifically, the portion of the App Sources section which is identical to the current 0.8.6 docs.

Having a large repeated content between the two versions feels distracting to me, especially on a feature which does have some significant and non-obvious changes. e.g., I skimmed that section, saw that it was mostly the same as 0.8.6 (and not copy-pastable, unlike the code blocks above), and ended up not noticing the part at the bottom where it said that you now have to explicitly add apps into the catalog.


gregelin:
@ethankaminski - The duplicate conference was intentional so that users could stay in the 0.9.0 documentation for getting started.

At some point, a user may need to go to the 0.8.6 documentation, but I wanted the user to consciously navigate away instead of providing a link and user getting confused between what is 0.8.6 and what is 0.9.0.

PR #667, opened 2019-07-17
adapt the 'docker_machine_aws.sh' HOST changes to 'docker_compose_nginx.sh'

peterkaminski:
tested and working:
* without `--address`
* with `--address test1.example.com`
* with `--address test1.example.com:1234`


gregelin:
@peterkaminski I made a couple of small tweaks. When no address was specified, $PORT caused an unbounded variable error. I also wrapped the report of the "Remember to update DNS" feedback in a conditional to only display if a domain name was specified.

PR #697, opened 2019-08-12
migration for organizations to portfolios

thejuliekramer:


siteapp/migrations/0030_auto_20190812_1907.py
```diff
@@ -0,0 +1,54 @@
...
+
+        for project in projects:
+            if project.organization and project.portfolio is None:
+                # get or create portfolio from project organization
+                portfolio, created = Portfolio.objects.get_or_create(title=project.organization.name)
+                # assign projects portfolio id
+                project.portfolio = portfolio
+                # save project
+                project.save()
+                # assign portfolio permissions
```
gregelin: I think the approach is correct, but I'm not sure if these role assignments are correct. Is there an easy way to list all the permissions that exist in Guardian?

thejuliekramer: every object has view, change, add and delete - you can also define custom permissions beyond that - owner was a custom permission I defined on portfolio for instance that allows someone to invite another person to have view permissions

siteapp/migrations/0030_auto_20190812_1907.py
```diff
@@ -0,0 +1,80 @@
...
+            # assign editor permissions
+            logger.info("assigning editor permission for {} to {}".format(pm.project, pm.user))
+            assign_editor_permissions(apps, project, pm.user)
+            if pm.is_admin:
+                # if admin assign owner permissions
+                logger.info("assigning owner permission for {} to {}".format(pm.project, pm.user))
+                assign_owner_permissions(apps, project, pm.user)
+
+    # TODO
+    # what to do with project.is_account_project?
```
thejuliekramer: do we need to consider these for the migration?

PR #700, opened 2019-08-14
Add object creation dropdown to navbar

gregelin:
Add a "+" object creation dropdown to navbar
to remove the need to navigate back to the
projects page or inside a portfolio to create
a new project or portfolio.

Add/Create a project or portfolio from anywhere.
Also, if we are within a portfolio, make sure the
default portfolio option that comes up is the
current portfolio.

modified:   guidedmodules/views.py
new file:   siteapp/tests.orig.py
modified:   siteapp/views.py
modified:   templates/base.html
modified:   templates/navbar.html
modified:   templates/portfolios/detail.html

### DIFF
--- a/guidedmodules/views.py
+++ b/guidedmodules/views.py
@@ -8,10 +8,12 @@ from django.db import transaction
 import re
 
 from .models import Module, ModuleQuestion, Task, TaskAnswer, TaskAnswerHistory, InstrumentationEvent
+
 import guidedmodules.module_logic as module_logic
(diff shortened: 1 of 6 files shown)


thejuliekramer:
👍 looks good, have all of the initial objects been set where possible?

gregelin:
@thejuliekramer Yes, I set the initial objects to the current portfolio on the pages where user was clearly within a portfolio.

PR #718, opened 2019-08-30
Ge fix perms proj and comments 0830

gregelin:
Fix Portfolio members editing Portfolio Projects

Append new Django Guardian permission check to legacy
0.8.6 permission checks to make sure users with
permissions to view, edit, change a portfolio can
view, edit, change all portfolio projects. In other words,
a Portfolio member can edit all Portfolio Projects.

Learned from Josh that we have many permissions on individual
tasks because we created tasks (and teir permissions) before
we created Projects (or Portfolios). So we have a lot of bottom-up
permissions management sprinkled throughout the code.

Now that we have Projects, and Django Guardian, it will make
more sense to check Project permissions rather than task permissions.
In future we can increase the granularity of the permissions back to
questions (tasks) and answers.

The fast fix is to append 0.9.0 Guardian permission check to
a legacy 0.8.6 check, which we do here.

We also fix a few typos in comments and HttpResponseForbidden message
to help debug.

Remove legacy Organization from discussion invites, turn off discussion polling

Non-super users were not able to submit discussion comments
because the invitations being sent upon comment add still included
legacy organization reference. Invitations no longer have Organizations
in model after the removal of multi-tenant sub domains.

Removing the reference to organizations solved the issue with users
not being able to add comments.

We won't need organizations on discussion in future either, because
discussions will be tied to projects (or tasks) that still have
organization structure built-in.

Also turned of javascript initiation of polling for updates on
discussions in order to remove that noise from log. Polling was
a cheap way to check.

Add column headers to Projects, Portfolios listing

### DIFF
--- a/guidedmodules/models.py
+++ b/guidedmodules/models.py
@@ -930,10 +930,10 @@ class Task(models.Model):
         return None
 
     def has_read_priv(self, user, allow_access_to_deleted=False):
-        return self.get_access_level(user, allow_access_to_deleted=allow_access_to_deleted) in ("READ", "WRITE")
+        return self.get_access_level(user, allow_access_to_deleted=allow_access_to_deleted) in ("READ", "WRITE") or self.project.has_read_priv(user)
 
     def has_write_priv(self, user, allow_access_to_deleted=False):
-        return self.get_access_level(user, allow_access_to_deleted=allow_access_to_deleted) == "WRITE"
+        return self.get_access_level(user, allow_access_to_deleted=allow_access_to_deleted) == "WRITE" or self.project.has_read_priv(user)
 
     def has_review_priv(self, user):
         if self.project.organization is None:
@@ -1836,7 +1836,6 @@ class TaskAnswer(models.Model):
             for user in self.task.project.organization.help_squad.all():
                 if user in self.get_notification_watchers(): continue # no need to invite
                 inv = Invitation.objects.create(
-                    organization=self.task.project.organization,
                     from_user=comment.user,
                     from_project=self.task.project,
                     target=comment.discussion,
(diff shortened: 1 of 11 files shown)


guidedmodules/models.py
```diff
@@ -930,10 +930,10 @@ def get_access_level(self, user, allow_access_to_deleted=False, recursive=True):
         return None
 
     def has_read_priv(self, user, allow_access_to_deleted=False):
-        return self.get_access_level(user, allow_access_to_deleted=allow_access_to_deleted) in ("READ", "WRITE")
+        return self.get_access_level(user, allow_access_to_deleted=allow_access_to_deleted) in ("READ", "WRITE") or self.project.has_read_priv(user)
 
     def has_write_priv(self, user, allow_access_to_deleted=False):
-        return self.get_access_level(user, allow_access_to_deleted=allow_access_to_deleted) == "WRITE"
+        return self.get_access_level(user, allow_access_to_deleted=allow_access_to_deleted) == "WRITE" or self.project.has_read_priv(user)
```
peterkaminski: since this is `has_write_priv()`, did you want this to be `self.project.has_read_priv()` or did you want `self.project.has_write_priv()`?

gregelin: Fudge. There's currently no `project.has_write_priv(user)` method available.

I think we merge the request and then create the segregated `has_write_priv`. I'm not sure how much more work there is to create that segregated privilege. I think we are being a bit lenient during the transition from 0.8.6 to 0.9.0 to give anyone with view permissions also edit permissions. The reason is we also have to create UI in order to set these permissions.

peterkaminski: Sounds good, I'll merge.  I made a placeholder issue to remind us to create `has_write_priv` and UI, when it's a good time: #719.

PR #722, opened 2019-08-31
Fixes #712. Check/set emailed status of notifications before sending.

gregelin:
Modify send_notifications_emails command to set a row lock on the
notification message in the database and make sure the notification
has not yet been emailed before we send the email. This change
should prevent to running instances of GovReady-Q in different
containers talking to the same database from each emailing
the same notification.

modified:   siteapp/admin.py
modified:   siteapp/management/commands/send_notification_emails.py

### DIFF
--- a/siteapp/admin.py
+++ b/siteapp/admin.py
@@ -4,6 +4,7 @@ from guardian.admin import GuardedModelAdmin
 import django.contrib.auth.admin as contribauthadmin
 
 from .models import User, Organization, Folder, Project, ProjectMembership, Portfolio
+from notifications.models import Notification
 
 def all_user_fields_still_exist(fieldlist):
     for f in fieldlist:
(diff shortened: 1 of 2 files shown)


JoshData:
The logic seams reasonable. You may want to document that [this will not work with Sqlite](https://docs.djangoproject.com/en/2.2/ref/models/querysets/#select-for-update), although it would be unlikely that a multi-instance Q would be running on a Sqlite database anyway, and it won't work with the older (currently non-default) [MyISAM](https://dev.mysql.com/doc/refman/5.7/en/myisam-storage-engine.html) MySQL database engine because it does not support transactions.

The only thing is now you have two fields tracking whether a notification has been sent, `Notification.emailed` and `User.notifemails_last_notif_id`. It might be possible to do this transaction thing solely on the `User.notifemails_last_notif_id` field so that you don't need to use `Notification.emailed` at all. Simpler logic is happy logic.

You could also consider writing a migration to set `Notification.emailed` on all past notifications (`Notification.id` <= `User.notifemails_last_notif_id`) and then drop `User.notifemails_last_notif_id` from the database entirely. But `User.notifemails_last_notif_id` may come in handy later. If a user turns off notification emails (=> `User.notifemails_enabled = false`) and then turns it on again later, you don't want to suddenly send emails for all of the notifications that were generated while the notification emails setting was off. When a user changes that setting from off to on, you can reset `User.notifemails_last_notif_id` to the highest id of a Notification in the database, to prevent any Notifications up to that point from getting emailed. So you could document the difference between the two ways you're tracking if a notification has been sent. (Alternatively, drop the `notifemails_last_notif_id` field from the database and when a user has `notifemails_enabled` turned off, just mark any notification as `emailed` to prevent it from being sent at a later time, even though it wasn't actually emailed.)

I don't think the last `save` is necessary - the transaction can close without you making any changes to the table, afaik.

Anyway, this PR looks reasonable, but I can't say for sure whether it's right. As with most concurrency things, there's no way to know if it's really doing what you're thinking until you try it in a multi-process setup and see that it actually solves the problem.

gregelin:
@JoshData Thanks for reviewing and your thoughts.

gregelin:
Tested this modification against two instances running against a single Postgres database and notifications were not emailed twiced.

PR #723, opened 2019-09-01
Ge fix helpers 0901

gregelin:
Fix help squad, reviewer settings management in 0.9.0.

Management of help squad and reviewers was tied to Organization in 0.8.6. Javascript calls for adding/removing users to help squad and reviewers assumed request passed organization value.

For initial fix, we stop looking to request for organization and set organization to the main organization.
    
Also add username to User.render_context_dict() method so we can autocomplete on username.

Tweak Organization and User Admin list to see id and username respectively in records list.

### DIFF
--- a/siteapp/views.py
+++ b/siteapp/views.py
@@ -22,7 +22,7 @@ from guidedmodules.models import (Module, ModuleQuestion, ProjectMembership,
 from .forms import PortfolioForm, ProjectForm
 from .good_settings_helpers import \
     AllauthAccountAdapter  # ensure monkey-patch is loaded
-from .models import Folder, Invitation, Portfolio, Project, User
+from .models import Folder, Invitation, Portfolio, Project, User, Organization
 from .notifications_helpers import *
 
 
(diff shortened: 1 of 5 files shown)


siteapp/views.py
```diff
@@ -1514,37 +1516,41 @@ def preload_profiles(users):
 def organization_settings_save(request):
     if request.method != "POST":
         return HttpResponseForbidden()
-    if request.user not in request.organization.get_organization_project().get_admins():
-        return HttpResponseForbidden()
+
+    # In 0.9.0 we only have 1 organization, so let's get that
+    # TODO better setting of organization
+    organization = Organization.objects.get(id=1)
```
JoshData: I would do something like `Organization.objects.get()`. If there is one in the database, it will return it. If there is zero or more than one, it will raise an exception. This way if the assumption is wrong, there's an error, rather than possibly editing the wrong Organization instance.

gregelin: Well, there is a chance in the future we could go back to having more than 1 organization. During install, the first organization created is the main organization.

JoshData: All the more reason to have it not silently take Organization 1.

PR #735, opened 2019-09-04
Require special env var to prevent accidental migrations to 0.9.0

gregelin:
Because Docker automatically runs database migrations, we add
the special environment variable DB_BACKED_UP_DO_UPGRADE that
must be intentionnally set to 'True' for dockerfile_exec.sh to
fully execute and migrate the database in deployments using Docker.

We test to make see if deployment's database is still in state
of version 0.8.6 and if so, then look for the DB_BACKED_UP_DO_UPGRADE
environment variable. If database has already been migrated, test
is ignored. If DB_BACKED_UP_DO_UPGRADE is set correctly to "True",
we proceed with migration.

This test avoids accidental migrations of databases without
organization first reading about upgrade to 0.9.0 and being
given the opportunity to back up database.

### DIFF
--- a/deployment/docker/dockerfile_exec.sh
+++ b/deployment/docker/dockerfile_exec.sh
@@ -90,7 +90,35 @@ echo "Starting at ${ADDRESS} with HTTPS ${HTTPS-false}."
 python3.6 manage.py check --deploy
 
 # Check if 0.9.0 upgrade has happened
-DB_BEFORE_090 ="$(python3.6 manage.py db_before_090_migration_check)"
+DB_BEFORE_090=$(python3.6 manage.py db_before_090)
+if [ $DB_BEFORE_090 = "True" ]
+then
+	echo "** WARNING!! **"
+	echo "Launching this container will automatically upgrade your GovReady-Q deployment to version 0.9.0!"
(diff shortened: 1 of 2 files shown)


deployment/docker/dockerfile_exec.sh
```diff
@@ -90,7 +90,35 @@ echo "Starting at ${ADDRESS} with HTTPS ${HTTPS-false}."
...
+			echo "'DB_BACKED_UP_DO_UPGRADE' environment variable not set to 'True'."
+			echo "To confirm you have backed up your database and deploy version 0.9.0, set the 'DB_BACKED_UP_DO_UPGRADE' environment variable to 'True' for your deployment."
+			echo "Launch and deployment halted to protect your existing database."
+			exit 1
+		else
+			echo "Confirmed 'DB_BACKED_UP_DO_UPGRADE' environment variable is set to 'True'."
+			echo "Continuing with deployment."
+		fi
+else
+	echo "Confirmed upgrade to version 0.9.0 previously completed."
```
peterkaminski: It would be nice if this were more specific about database migration.  How about:

"Confirmed that database has been migrated, and is compatible with version 0.9.0."

PR #743, opened 2019-09-10
HOLD Till 0.9.0. merge to master - 0.9.0.dev remove uneeded migration

thejuliekramer:
The 'view_portfolio' and 'view_project' permissions are being removed in this pull request because when the master branch is upgraded to the latest version of Django (2.2) the 'view' permission in all models comes by default. Because 0.9.0 will be merged after the upgrade not removing these parts of the migration will cause a duplicate error.


gregelin:
@thejuliekramer Could you please add a bit more detail about the permissions that are being removed and why? Also, is this preliminary to or concurrent with the Django upgrade? (If continuous, I'm wondering if the changes in `requirements.in` should be part of PR.

PR #749, opened 2019-09-29
Ge master synch

gregelin:
Pushing this PR because local `master` branch is reporting 3 commits ahead of `origin/master` but I can't find differences when using `git diff`. Attempting to see what GitHub says about the differences...


gregelin:
Because there are no file changes, I am self-merging as an administrator.

PR #758, opened 2019-10-11
update Python's SQLite to 3.8.5

peterkaminski:
Django 2.2 requires SQLite 3.8.3 or later. Standard CentOS has 3.7.17.

Python loads SQLite from the shared library in /lib64.

This PR updates the SQLite shared library version to 3.8.5, from a CentOS Atomic Host rpm.

The `fetch-vendor-resources` tool is used, so the download checksum can be verified.

I verified that the docker image gets built and runs.


peterkaminski:
Darn, the CircleCI build failed because `yum` is called in `fetch-vendor-resources.sh`, even though it worked for me.

```
./fetch-vendor-resources.sh: line 58: yum: command not found
Exited with code 127
```

peterkaminski:
The SQLite problem in CircleCI build is fixed.  The problem was that CircleCI is (currently) building in Debian Jessie, which of course doesn't need the same SQLite upgrade as CentOS 7.  The `fetch-vendor-resources.sh` script now checks that it is running under CentOS 7 before attempting the SQLite upgrade.

So, we're on to a new CircleCI failure, which happens during Selenium testing:

```
selenium.common.exceptions.ElementClickInterceptedException: Message: element click intercepted: Element <li id="tab-signin" role="presentation">...</li> is not clickable at point (847, 43). Other element would receive the click: <div id="navbar-collapse-target" class="collapse navbar-collapse">...</div>
  (Session info: chrome=75.0.3770.100)
```

Offhand, I don't know why this PR would elicit this bug any more than master would.

peterkaminski:
The Selenium problem disappeared after merging in the version bump from master.  So apparently it was a CircleCI hiccup, or the Selenium test fails intermittently.

PR #759, opened 2019-10-11
bump version to '0.9.0' (release)

peterkaminski:


peterkaminski:
CircleCI build failed with:

```
python3 .circleci/check_version_file.py
ERROR: CHANGELOG has content before a version heading while VERSION file does not include '+devel'. For version releases, VERSION should not include '+devel' and no information should come before the version heading in CHANGELOG. Alternatively, the version number 'v0.9.0' should end with +devel to signal that this is a development build.
```

PR #760, opened 2019-10-15
Change Portfolio.title max_length to 255

peterkaminski:
Avoids the following problem:

```
ERRORS:
siteapp.Portfolio.title: (mysql.E001) MySQL does not allow unique CharFields to have a max_length > 255.
```

Also changed from 256 to 255 (for consistency only, they're not unique):

* Organization.name
* Folder.title
* Invitation.to_email


gregelin:
Cloud.docker.com is currently unavailable, effecting builds.

PR #772, opened 2019-11-15
Allow use of alternative virtual environments.

YPCrumble:
@gregelin @cityinohio2019 nice to e-meet you. This is Ian Campbell - we emailed today / I'm speaking with Dayton tomorrow. I just started playing around a little with the platform using the local development tutorial and thought to make a couple (super tiny) contributions to see how things worked as well as check out the CI environment.

This PR allows users to use their own virtualenvironment setup vs. forcing the virtualenv to be part of the working directory.

Feel free to close if this isn't helpful - I'm just testing things out!

Checks for the presence of $VIRTUALENV_ENV to detect an existing
virtual environment, per https://stackoverflow.com/a/28388115/2532070.

If a virtualenvironment exists, install requirements to this
environment.

### DIFF
--- a/quickstart.sh
+++ b/quickstart.sh
@@ -33,7 +33,7 @@ then
 	pip3 install -r requirements.txt;
 elif [ -v $VIRTUALENV_ENV ]
 then
-  pip3 install -r requirements.txt;
+	pip3 install -r requirements.txt;
 else 
 	pip3 install --user -r requirements.txt;
 fi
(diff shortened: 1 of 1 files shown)


quickstart.sh
```diff
@@ -31,7 +31,10 @@ if test -f env/bin/activate
 then
 	source env/bin/activate;
 	pip3 install -r requirements.txt;
-else
+elif [ -v $VIRTUALENV_ENV ]
+then
+  pip3 install -r requirements.txt;
```
gregelin: @YPCrumble Could you adjust the line indent to be further in for consistency?

YPCrumble: @gregelin ah thank you!  My editor was set so that I didn't realize this had tabs vs. spaces. Just updated and removed spaces lower down in the file for consistency. Let me know if that isn't what you were looking for and I can revert adjust as appropriate. 

Another option would be to convert this to spaces as it seems most of the rest of this repo project doesn't use tabs, though I didn't do that as it would make inspecting the commit history more difficult.

Looking forward to speaking with you today!

YPCrumble:
@gregelin you mentioned:

> And now you also have me wondering about the line before: there seems to be an assumption of as to the local path to the virtual environment. So I think you found a spot where more refactoring is needed.

I left the line there as a convenience function for users who use the `virtualenv -p python3 env` command [mentioned in the quickstart docs here](https://govready-q.readthedocs.io/en/latest/deploy_local_dev.html#quickstart). I just wanted to be able to use my own virtualenv setup and still use the `quickstart.sh` script.

PR #774, opened 2019-11-17
0.9.0.dev Performance improvements

gregelin:
* Create migrations for shortening varchar field length

* Identify current_group in progress-history when id and module-id differ. Fixes an issue where the `id` and `module-id` of an app.yaml question differ caused the progress history to fail display questions properly in `task-progress-history-list.html` template.

* To dramatically improve performance, set `is_finished` check to False when showing question or showing module review page. Current `is_finished` exhibited poor performance causing caches to be rebuilt when answers changed on existing questions. We were checking `is_finished` when building the `task-progress-history-list`. We need a better `is_finished` test to answer state quickly. Only lost on user interface is not seeing a check mark for an entire
module in the `task-progress-history-list`.


guidedmodules/views.py
```diff
@@ -644,12 +650,18 @@ def render_markdown_field(field, output_format, **kwargs):
             if mq.spec.get("protocol"):
                 can_start_any_apps = True
 
-        if questions.get(mq.id) and 'task' in questions[mq.id]:
-            # print("\nquestions[mq.id]['task'].id", dir(questions[mq.id]['task']))
+        # if questions.get(mq.id) and 'task' in questions[mq.id]:
+        # print("** questions.get(mq.id)", questions.get(mq.id)) # Debug
+        if 'task' in questions[mq.id]:
```
peterkaminski: Why this change?  Will `questions.get(mq.id)` always exist?

gregelin: Line 653 and 654 were debug lines. I know this because if you look at Line 908 you will see the code repeated. `questions.get(mq.id)` will always reply with an id value or None and if None, the else condition is executed.

peterkaminski: if `questions[mq.id]` is None, won't you get a `TypeError: argument of type 'NoneType' is not iterable`?

gregelin: I answered your previous question poorly, and have edited my answer.

`questions.get(mq.id)` was part of a debug effort. `mq.id` is always defined within this code block.

gregelin:
@peterkaminski Something else may be happening here.

I am running the tests with code prior to 56c097c, and running just the invitation test.  

The `NoneType` is coming from a setup function `_start_task` (L300) that has a failing regex (L308). `test_invitations` fails because the is calling `_start_task` errors.  function. (NOTE: Line numbers are from prior my debug code.)

I added in a debug `print("self.browser.current_url: '{}'".format(self.browser.current_url))` line around `tests.py` L308 that yields output:

```
self.browser.current_url: 'http://localhost:60171/projects/3/i-want-to-answer-some-questions-on-q'
```

Notice the output URL contains `project` rather than `task` as the test setup function expects.


Here is complete execution of `python3 manage.py test siteapp.tests.GeneralTests.test_invitations`

```bash
Gregs-MacBook-Pro:govready-q-0.9.0-module-nav greg$ python3 manage.py test siteapp.tests.GeneralTests.test_invitations
Creating test database for alias 'default'...
System check identified no issues (0 silenced).
** executing evaluate_module_state - 
** finish executing evaluate_module_state - 
self.browser.current_url: 'http://localhost:63283/projects/3/i-want-to-answer-some-questions-on-q'
E
======================================================================
ERROR: test_invitations (siteapp.tests.GeneralTests)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/codedata/code/govready-q-0.9.0-module-nav/siteapp/tests.py", line 444, in test_invitations
    self._start_task()
  File "/codedata/code/govready-q-0.9.0-module-nav/siteapp/tests.py", line 309, in _start_task
    return Task.objects.get(id=re.search(r"/tasks/(\d+)/", self.browser.current_url).group(1))
AttributeError: 'NoneType' object has no attribute 'group'

----------------------------------------------------------------------
Ran 1 test in 50.596s

FAILED (errors=1)
Destroying test database for alias 'default'...
```

gregelin:
@peterkaminski I just deleted an earlier comment regarding error in `_start_task`. I  was getting different results due to configuration of static files location.

PR #814, opened 2020-03-21
Fix appsource field error handling for Django 2.2

gregelin:
This commit fixes two problems to gracefully handle appsource
field validation errors when updating an appsource and adding
incorrect data.

One error was appsource page trying to render the widget for
for available apps where no available apps existed because
an appsource connection could not be made.

The other error appeared to be failing to render error messages
properly through Django forms (for admin) because we needed to json.dumps
a variable that was an OrderedDict and was previously passed without
dumping to a string formatted as JSON.

### DIFF
--- a/guidedmodules/admin.py
+++ b/guidedmodules/admin.py
@@ -1,6 +1,7 @@
 from django.contrib import admin
 from django import forms
 from django.utils.html import escape as escape_html
+from django.contrib import messages
 
 from .models import \
 	AppSource, AppVersion, Module, ModuleQuestion, ModuleAsset, \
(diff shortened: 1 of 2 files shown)


guidedmodules/admin.py
```diff
@@ -1,6 +1,7 @@
 from django.contrib import admin
 from django import forms
 from django.utils.html import escape as escape_html
+from django.contrib import messages
```
JoshData: This line doesn't seem to be needed. Or if it fixes a different issue, it's not mentioned in the commit message.

gregelin: Line removed.

PR #816, opened 2020-03-24
Gracefully render blank datagrid in output templates

gregelin:
When a datagrid question is not answered we still need to be able to render the output template referencing datagrid question without crashing.

### DIFF
--- a/guidedmodules/module_logic.py
+++ b/guidedmodules/module_logic.py
@@ -711,7 +711,18 @@ class HtmlAnswerRenderer:
             # Assuming that RenderedAnswer gives us string version of the stored datagrid object
             # that is an Array of Dictionaries
             import ast
-            datagrid_rows = ast.literal_eval(value)
+            try:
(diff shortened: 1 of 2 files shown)


guidedmodules/module_logic.py
```diff
@@ -711,7 +711,18 @@ def __call__(self, question, task, has_answer, answerobj, value):
...
+                datagrid_rows = ast.literal_eval(value)
+            except:
+                if value == "<nothing chosen>":
+                    # Datagrid question has been visited and instantiated but no answer given
+                    # No data was entered into data grid
+                    datagrid_rows = []
+                else:
+                    # Datagrid question has not been visited and not yet instantiated
+                    # `value` is set to "<Software Inventory (datagrid)>"
+                    datagrid_rows = []
```
peterkaminski: Did you want a value here, or really `[]`?

gregelin: For now, yes.

PR #833, opened 2020-04-20
Control Guidance using OSCAL

gregelin:
Incorporate better control guidance with OSCAL. Initial development of control "Statement" model and component to control mapping.


templates/navbar.html
```diff
@@ -21,6 +21,7 @@
           <li><a href="/projects" id="menu-projects">Projects</a></li>
           <li>&nbsp;&nbsp;&nbsp;&nbsp;</li>
           <li><a href="/portfolios" id="menu-portfolios">Portfolios</a></li>
+          <li><a href="/controls/800-53//" id="menu-controls">Controls</a></li>
```
peterkaminski, line 24: Is the double slash needed or not? `href="/controls/800-53//"`

gregelin: A modification and update has addressed this issue. The "//" was incorrect. Fixed routing and added a better temporary index page for controls.

PR #842, opened 2020-05-02
Make PDF generation and use of wkhtmltopdf optional

gregelin:
Make PDF and image generation optional and configurable.

- Individuals can still download Word documents
- The PDF Generation software, wkhtmltopdf, had other potential security issues in which malicious users could with some effort get wkhtmltopdf to blindly include server side files within the generated PDF.

PDF generator library wkhtmltopdf has security issues wherein individuals could add
HTML references such as links or file references inside the documents
they are creating which the PDF Generator blindly interpets. This leads
to SSRF (Server Side Request Forgery) in which data is retrieved from
server and added to PDF by the PDG Generator. We are also encountering
issues with the sub-dependency of `libxslt` on older versions of CentOS
before CentOS 8.x raising CVE vulnerability issues with scanners. For these
reasons, PDF Generation is being a configurable setting.

Supported options are currently 'None' and 'wkhtmltopdf' for PDF and
'None' and 'wkhtmltoimage' for image generation. This leaves
existing `wkhtmltopdf` code in place while creating default to 'None'.

If no PDF generator is set, the option to download PDF is removed from
task_finished page. As a backup, if there is an attempt to generate
a PDF from the Tasks model, the method gracefully fails when no
PDF generator is set and produces a text file instead with a message
stating PDF is not configured.

If `wkhtmltoimage` or valid image generator is not available, thumbnails
cannot be generated for uploaded files, either. Gracefully avoid trying
to create a thumbnail

Support new `environment.json` parameters of `GR_PDF_GENERATOR`
and `GR_IMG_GENERATOR`.


gregelin:
More testing is needed.

- [ ] test with environment settings turning on PDF generation
- [ ] test with/without in docker
- [ ] need to test uploading a file to a question

gregelin:
@peterkaminski 
Bandit does not like replace of `/usr/bin/pandoc` with just `pandoc`. Do we know how to get Bandit to accept this? Or do we change it back?

```bash
Test results:
 
>> Issue: [B607:start_process_with_partial_path] Starting a process with a partial executable path
 
   Severity: Low   Confidence: High
 
   Location: guidedmodules/models.py:1207
 
   More Info: https://bandit.readthedocs.io/en/latest/plugins/b607_start_process_with_partial_path.html
 
1206	                outfn = os.path.join(tempdir, filename)
 
1207	                with subprocess.Popen(
 
1208	                    ["pandoc", "-f", "html", "-t", pandoc_format, "-o", outfn],
 
1209	                    stdin=subprocess.PIPE
 
1210	                    ) as proc:
 

 ```

gregelin:
Successfully told Bandit to ignore the pandoc issue.

PR #870, opened 2020-05-27
Correct setup of tests in discussions app

gregelin:
Make sure setup in discussion test is the same as setup in
siteapp to avoid failing due to missing system modules.


gregelin:
Much needed fix in testing done with Josh to resolve a failure due to database not being set up correctly for tests.

PR #898, opened 2020-05-30
Update Ubuntu from source install docs

gregelin:
Ubuntu from source installation docs updated to include upgrading
pip to solve problem of Ubuntu's defualt pip not installing
Python packages under Linux user installs.


gregelin:
Documentation change only.

PR #903, opened 2020-06-02
Include Xacta Excel Export

gregelin:
Add link to export selected controls for system to an excel spreadsheet using the Xacta spreadsheet import format. Currently does not separate common controls from other controls in rows.


controls/views.py
```diff
@@ -122,6 +122,271 @@ def controls_selected(request, system_id):
...
+            stream = tmp.read()
+            blob = stream
+
+        mime_type = "application/octet-stream"
+        filename = "{}_control_implementations.xlsx".format(system.root_element.name.replace(" ","_"))
+        
+        resp = HttpResponse(blob, mime_type)
+        resp['Content-Disposition'] = 'inline; filename=' + filename
+        return resp
+
```
peterkaminski, line 378: Lines 378-385 are after the `return` statement at line 377.  Is that what you want?

PR #918, opened 2020-06-05
Improve Ubuntu from source docs HTTPS instructions

gregelin:


gregelin:
Changes are documentation only.

PR #923, opened 2020-06-07
Update ports in NGINX https example to be simple 80, 443

gregelin:
Need to more research to  understand 
listen [::]:80; and listen [::]:443; format. So dropping back to simpler listen 80; and listen 443; format.


gregelin:
Documentation change only.

PR #925, opened 2020-06-07
CRITICAL fix for Ubuntu deploy docs and config files.

gregelin:
Remove all comments from `supervisor-govready-q.conf ` because
trailing "# comments" cause line to be ignored. This caused a problem
in Ubuntu deployment documentation leading to a situation where
gunicorn was starting in wrong context and python packages not found.

Improved `settings.py` handling of deprecated `host` and `https`
parameters for setting `SITE_ROOT_URL` value.


gregelin:
Change was primarily documentation plus an urgent change in handling environment variables in `siteapp/settings.py`.

Issue #934, opened 2020-06-12
Infinite redirects with SSL HTTPS behind NGINX reverse proxy depending on configuration


gregelin:
Combination of `SECURE_SSL_REDIRECT = True` and NGINX reverse proxy passing http rewrites can lead to infinite redirects. 

This is [noted in Django documentation](https://docs.djangoproject.com/en/3.0/ref/settings/#secure-ssl-redirect).

> If turning this to True causes infinite redirects, it probably means your site is running behind a proxy and can’t tell which requests are secure and which are not. Your proxy likely sets a header to indicate secure requests; you can correct the problem by finding out what that header is and configuring the SECURE_PROXY_SSL_HEADER setting accordingly.

Problem occurs when `local/environment.json`'s `govready-url` includes scheme `https://` (or deprecated `"https": true`) **and** NGINX has a redirect setting for port 80 that includes passing the `$request_uri` portion of the URL. It is necessary to include `$request_uri` in the redirect to pass invitations from `HTTP` to `HTTPS`.

Options to fix include setting a new in `local/environment.json` parameter:

* `"proxy": true` to make sure that `SECURE_SSL_REDIRECT` is never set True behind a proxy.
* `"secure_ssl_redirect": false` to force setting of `SECURE_SSL_REDIRECT`
* `"secure_ssl_redirect": true` to force setting to true when DJANGO when you want DJANGO to manage the redirect.

gregelin:
It might be best to set `SECURE_SSL_REDIRECT = False` in `siteapp/settings.py` and permit an override from `"secure_ssl_redirect": true` in `local/environment.json` file. 

In non-proxy environments, `HTTPS` would work with `SECURE_SSL_REDIRECT` set to false, and `HTTP` settings would fail as might be expected. Users that want to redirect could search the documentation and learn of the option to redirect.

Also, in proxied environments, the redirect would be managed by the proxied environment. 

And the meaning of the scheme `https://` in `govready-url` parameter (or deprecated `"https": true`) would continue to be accurate.

gregelin:
Changes to make:

- [ ] Update `siteapp/settings.py` to conditionally look for `secure_ssl_redirect` parameter in django's environment variables
- [ ] Add comment to `siteapp/settings.py` explaining setting
- [ ] Update installing GovReady-Q instructions. Be sure to include a note on infinite redirects.
- [ ] Alert colleagues

PR #936, opened 2020-06-12
Remove unneeded SECURE_SSL_REDIRECT

gregelin:
I had an extra `SECURE_SSL_REDIRECT = False` that was not needed.


gregelin:
Tested locally. Passed.

gregelin:
This is small follow on correction to #934

PR #945, opened 2020-06-15
Display statement counts w selected controls page; add datetime to project export

gregelin:
Display component statement counts on a system's selected controls page via a custom tag to provide access to a dictionary key of impl_smts_counts created in the view for the page.

Improve export project filename with datetime.

Change Org field to select in django projectadmin.


gregelin:
Tested locally.

PR #947, opened 2020-06-15
Dynamically determine the control_catalog to use in template

gregelin:
Dynamically determine the control_catalog to use in template
based on the oscal_catalog_key of the first control in selected_controls.
This makes the identification of the control_catalog automatica and
allows dynamic switching of control catalog content in output templates
but has a short coming of only choosing one catalog.

TODO: Retrieve multiple catalogs because we could have catalogs plus overlays
Will need a better way to determine the catalogs on a system so we can retrieve at once
Maybe get the catalogs as a property of the system

* Fix small typos in NIST 800-171 Rev 1 catalog

* Display app version in small app listings in store


gregelin:
Tested locally and all tests pasted.

Issue #953, opened 2020-06-16
Make sure non-admin users with permission to a project also have permission to the system


gregelin:
When a non-admin user is given access to a project, make sure non-admin user is also granted appropriate view and edit permissions for associated system and system.root_element.

When view permissions are not correctly assigned to a non-admin user to a project's related system and system.root_element, the non-admin user is unable to see the controls for the system (or assign controls for the system) because they do not have access to the page.

Implementing...

- [ ] Write tests
- [x] Identify permissions to add
- [x] Add code
- [x] Test in shell
- [x] Add in logging

gregelin:
Tests are proving challenging to write.

Issue #951, opened 2020-06-16
Assign permissions on system, root_element to non-admin creator of a new project


gregelin:
When a new project is created by a non-admin, correctly assign the permissions to the system and root_element to the user who started the project.

When ownership permissions are not correctly assigned to a non-admin user, the non-admin user is unable to see the controls for the system (or assign controls for the system) because they do not have access to the page.

Implementing...

- [x] Write tests
- [x] Identify permissions to add
- [x] Add code
- [x] Test in shell
- [x] Add in logging

gregelin:
Addressed in df4e2468397202c2d6b9a7f992db898a5d49a089.

PR #955, opened 2020-06-17
Assign permissions on invite fixes

gregelin:
Further fixes for #951 and #953 to make sure permissions assigned when a new user accepts invitation to a project.


gregelin:
Tested locally, too.

PR #965, opened 2020-06-24
Track status of component-control statement

gregelin:
Add field to track status of a statement, e.g.,
Planned, Implemented, Not Implemented, etc.

Also fix a couple incorrect comments on statement model fields.


gregelin:
Tested locally and all tests passed.

PR #966, opened 2020-06-24
Add/edit implementation statements on system component detail page

gregelin:
Include the ability to edit implementation statements and add new
implementation statements for a system component on the system
component view. This is nice because you manage the content
associated with a component on the component page for a system.


gregelin:
Tested locally and all tests passed.

PR #975, opened 2020-07-09
Display update date of control narratives

gregelin:
Add smts_updated field to ElementControl to store
date of a most recently updated control implementation
statement. Update views.save_smt to also update
ElementControl.smts_updated whenever statement is saved.

Display smts_updated value in system's selected_controls view.

Create a page where we can easily view recently updated
statements in reverse chronological order. Create tests for
the page first before we create the page.

Fix models Statement, Element that were not properly updating `updated` field
with current date when record changed.


gregelin:
All tests passed locally.

PR #979, opened 2020-07-17
Fix #839. Prevent case-insentive duplicate portfolio names.

### LINKED ISSUE #839: All Portfolios must be of a unique Title
Found by testing team.

"Title text box seems to be case sensitive. It allows the User to create a portfolio with an existing Title if the letters are of a different case than the existing one. 
E.g. TEST2 and Test2"

gregelin:
Make sure duplicate portfolio names are not created, even if duplication is case insensitive.

Fixes #839


gregelin:
All code tested locally. New tests written for use case.

PR #985, opened 2020-07-27
Add "part" attribute to Statement model

gregelin:
Add "part" attribute to Statement model to support tracking
control implementation statements by control parts.

Update the control implementation editor to support part information.

TODO:
- Add part information into generated OSCAL, OpenControl
- Make sure part information is displaying in SSP output


gregelin:
Problem with circleci checker has to do with version contradictions of importlib-metadata. But importlib-metadata  is a library to access the metadata for a Python package and not related to running code. It is intended to be ported to Python 3.8, but has not yet been imported.

PR #1002, opened 2020-08-13
Fix typo in statment permission edit denied logging

gregelin:
This fixes a small typo in referencing the statement.id when logging failed permissions to edit a statement.


gregelin:
Tested locally and tests passed.

Issue #1018, opened 2020-09-09
Incorporate rendering of SSP in OSCAL format


gregelin:
As a user, I want to be able to download an SSP in OSCAL formats (XML and JSON).

- [x] Display title of the template in task-finished instead of ID of template
- [x] Rename `oscal_template.json` to `ssp_v1_oscal.xml` or something similar
- [x] Read `format` field from template YAML to determine if an option should be shown for OSCAL

gregelin:
Current work is on branch `ge/oscal-ssp-0909`.

To see SSP generation, it is necessary to be using an compliance app that has an output template with format `oscal_xml`.
Such an example exists in [Lightweight_ATO_template](https://github.com/GovReady/Lightweight_ATO_template/tree/master/apps/light-ato-apps) (but currently on the [pk-oscal-template-20200824 branch](https://github.com/GovReady/Lightweight_ATO_template/tree/pk-oscal-template-20200824/apps/light-ato-apps)

gregelin:
Being addressed in PR #1026.

gregelin:
After more improvements, better generation of OSCAL in JSON now working and in release 0.9.1.48.

PR #1026, opened 2020-09-17
Improvements to support generating SSPs in OSCAL

gregelin:
Improve downloading of OSCAL output templates. Add support for downloading oscal in XML and JSON.
    
    See issue https://github.com/GovReady/govready-q/issues/1018
    for additional information.
    
    Improve the UI on the task-finish page to display option to download OSCAL versions of SSP.
    
    First, create new display of generated documents listed
    in a table format with different download options.
    This allows for only showing option to download OSCAL version
    when template is strictly for OSCAL. This display also makes
    it clearer what documents have been generated than when every
    document is in its own accordian. (When each output document
    was displayed within an accordian, it is not clear to user
    what documents are available.) Still showing accordians at this
    time, but they are lower on the page.


gregelin:
Implements issue #1003

PR #1029, opened 2020-09-20
Ge/reuse 0903

gregelin:
[WIP] Adding component reuse capability

This commit begins the work of enabling easy reuse of control-implementation
statements among multiple systems.

Supporting reuse of control-implementation statements (allocated by component)
is an essential part of Compliance as Code and accelerating and automating
the ATO/compliance process.

To reuse control-implementation statements authored in GovReady,
we need a way to capture a control-implementation statement when written and
make that statement available and easily found to be "copied" to other systems.

This commit adds a "prototype" attribute to statements to associate a given
statement with a "prototype" of that statement. We can now create statement_types
that have a suffix "_prototype" to identify a statement as an abstract, original
upstream statement prototype that can have individual instances assigned to system
element mappings.

We add two methods to the Statement model. We add `create_prototype` method to
make a deep copy of an existing statement and create a prototype of that statement.
We add `create_instance_from_prototype` method to do a deep copy of a prototype
statement and make the statement an instance. We link the instance of a statement
to the prototype statement using the `prototype` attribute.

Prototype control implementation statements have the type "control_implementation_prototype" and
are associated with a producer_element element. This makes it easy to identify all
the prototype statements for an abstract producer_element. We can also easily find all
the instances of a prototype statement across all systems by finding statements that
have a prototype relationship to our given prototype statement. This should also
make it very easy to diff instances of a statement against the prototype statement and
update accordingly.

As part of this commit, we add small form to the selected component's page of a system.
Here, the user can assign a new component to system via a drop down and the
`add_system_component` view will loop through control_implementation_prototype statements
associated with the component/element and add an instance of the statement to the element
if the control has been selected for the system.

The form for adding a system element to selected components uses Autocomplete via a library
called "select2" (see: https://select2.org/). To support the Select2 autocomplete library
a `css/select2.min.css` and `js/select2.min.js` have been added to static files. In might
be better in the future to include these in the `fetch-vendor-resources.sh` script.
In summary, a control implementation statement can be written for a system element and
turned into a control implementation prototype for that system element after which the
system element can be added to other systems and those systems receive copies of the controls.

Eventually we will report updates on the controls to the individuals system so they can
choose to inherit the controls.


gregelin:
This PR is superseded by https://github.com/GovReady/govready-q/pull/1111.

The branch `autocomplete_statements_#1066` is now the primary feature branch for functionality in `ge/reuse-0903` and `automated-tests-statements`

PR #1037, opened 2020-09-29
catching some windows-specific adapations to run govready locally. Ad…

davidpofo:
…ding pip-tools and python-magic-bin(fills in for the magic library).


gregelin:
@davidpofo This PR has been around for a while. Still relevant? Should branch be re-synched and tested against `develop`?

davidpofo:
Still relevant but not worth synching at the moment.

PR #1039, opened 2020-10-02
Fix navbar settings icon

gregelin:
Fix a quote that accidentally went missing and visually hid the settings icon shown to admins in the navbar menu.


gregelin:
This is a very small fix in a template. Code tested locally.

PR #1054, opened 2020-10-20
Update lxml version from 4.5.2 to 4.6.1 latest

bama4:
Updates lxml version from 4.5.2 to 4.6.1 in `requirements.txt`

PR #1055, opened 2020-10-21
Fixed loose file permissions for /tmp/environment.json file

bama4:
Changed file permissions of `/tmp/environment.json` to 0600


bama4:
`pyup.io/safety-ci` will pass once #1054 is merged.

gregelin:
@JoshData @peterkaminski There's a request to modify our Docker file. I believe it is a request to run Docker commands as `root`. Is this request OK to merge?

JoshData:
Some docker environments force the run time user to be a non-root user, and I think we were encouraged to set up containers in a manner that avoided using the root user, so I think this might break those deployments maybe?

bama4:
> Some docker environments force the run time user to be a non-root user, and I think we were encouraged to set up containers in a manner that avoided using the root user, so I think this might break those deployments maybe?

I fixed it. It now uses the `application` user to change the file permissions

gregelin (approved):
Looks good.

PR #1056, opened 2020-10-23
Need element control catch and escaping

davidpofo:


bama4:
`pyup.io/safety-ci` will pass once #1054 is merged.

PR #1088, opened 2020-11-10
Add organizational parameter value substitution.  [MTC-53]

woodt:


controls/models.py
```diff
@@ -58,7 +57,7 @@ def catalog_control_as_dict(self):
         """Return the control content from the catalog"""
         # Get instance of the control catalog
         catalog = Catalog.GetInstance(catalog_key=self.sid_class)
-        catalog_control_dict = catalog.get_flattended_controls_all_as_dict()
```
gregelin, line 61: Grateful for typo fix.

gregelin:
@woodt I like the inclusion of the organizational defined parameters as part of instantiating the Catalog object. It is accurate to think of a catalog as having been fully defined by the including an org's parameter.

I think would probably be an easy way for us to instantiate and cache different Catalogs for different organizations, so we would get the performance while still having Catalogs for each org in an instance. (I wonder how we determine memory requirements for that).

controls/oscal.py
```diff
@@ -54,15 +53,25 @@ class Catalog (object):
...
+
+        catalog_instance_key = '_cached_instance_' + catalog_key
 
-    def __init__(self, catalog_key='NIST_SP-800-53_rev4'):
+        if not hasattr(Catalog, catalog_instance_key):
+            new_catalog = Catalog(catalog_key=catalog_key, parameter_values=parameter_values)
+            setattr(Catalog, catalog_instance_key, new_catalog)
+        return getattr(Catalog, catalog_instance_key)
+
+    def __init__(self, catalog_key='NIST_SP-800-53_rev4', parameter_values=dict()):
```
bama4, line 74: `NIST_SP-800-53_rev4` could be a constant instead of the hardcoded string

controls/tests.py
```diff
@@ -200,13 +200,22 @@ class Oscal80053Tests(TestCase):
...
+        cg_flat = cg.get_flattened_controls_all_as_dict()
         control = cg_flat['au-2']
         self.assertEqual(control['id'].upper(), "AU-2")
         # self.assertEqual(control.class, "NIST.800.53")
         # TODO: ADD Class into object
         self.assertEqual(control['title'].upper(), "AUDIT EVENTS")
 
+    def test_catalog_controls_with_organizational_parameters(self):
+        parameter_values = { 'ac-1_prm_2': 'every 12 parsecs' }
+        cg = Catalog.GetInstance('NIST_SP-800-53_rev4')
```
bama4, line 212: `NIST_SP-800-53_rev4` could be a constant

controls/oscal.py
gregelin: Agreed @bama4

gregelin:
@woodt This is looking very good. Glad to see the tests in place.

gregelin:
@woodt We might want to address the database model for `Org_Catalog_Parameters` as a table with fields:

- organization
- catalog_key
- parameter_id
- label
- default
- default_type

If we could make the `organization` field arbitrary object type (as in attaching tags to different objects), then we could have one table for parameters whether related to organization, system, or portfolio.

controls/oscal.py
woodt: Made the catalog keys constants.

controls/tests.py
woodt: Made catalog keys constants.

woodt:
Cleaned up the API a bit by more closely tying org params to a catalog instance so the params don't need to be passed around
as much.  This now works because I changed the catalog instance cache to take into account the org parameters.  Added some basic persistence for organizational settings (we can add the additional fields Greg suggested when we add the features that require them.)  Removing the "draft" attribute from this PR.  I think it is ready for final review.

gregelin (approved):
Looks good. Code examined closely. Tests run. Small tweaks made.

Issue #1094, opened 2020-11-13
Missing current group in demo appsource


davidpofo:
As a user, when creating a new project with the demo appsources I should be able to see if there is a group for a question shown in the breadcrumbs. If there is no group then the breadcrumb should reflect that by omitting the group section.

PR #1095, opened 2020-11-13
Omitting current_group breadcrumb if it is None for the given question

davidpofo:
This address issue #1094


gregelin:
@davidpofo What about the case of no `current_group`? Should there be an `elseif` portion to your test?

gregelin (commented):
@davidpofo What about the case of no current_group? Should there be an elseif portion to your test?

davidpofo:
> @davidpofo What about the case of no current_group? Should there be an elseif portion to your test?

When there is no group assigned to a question it return None as is default earlier for the function `show_question` in the guidedmodules view return `current_group` this script. So instead of a group name it is just None like attached. My conditional is to include the rest of the breadcrumb but avoid producing a None in the path that will confuse users. Is there a default group you were thinking to substitute the None? If so it would be changing that default not adding another conditional in the template.

image

gregelin:
@davidpofo Thanks for the screenshot. The fix makes sense.

gregelin:
@davidpofo Can you please add entry to `CHANGELOG.md` and bump `VERSION`?

peterkaminski (approved):
lgtm

davidpofo:
Reviewed during PR party #2

PR #1098, opened 2020-11-15
Add names to testmocking/names.txt

gregelin:
Increase the number of unique names in `testmocking/names.txt` to be able to create more test users


davidpofo (approved):
Names are interesting but fine

Issue #1097, opened 2020-11-14
Fix the Fortify Password Management: Hardcoded Password False Positives


bama4:
**Description**

To avoid having to explain future Fortify issues that are false positive. The preferable solution is to fix some of the false positive findings such that they will not appear in future Fortify runs.

This ticket is for fixing the false positives under the `Fortify Password Management: Hardcoded Password` category.

**Goal**
- [ ] Fix all of the false positives that can be changed that fall under the `Fortify Password Management: Hardcoded Password` category.

**Code Practice Suggestions**
In the future, avoid using variables containing the word `PASSWORD`.

gregelin:
@bama4 Can we close this issue? Why is issue blocked?

bama4:
Finished, findings either explained or fixed

PR #1100, opened 2020-11-17
Changed fortify false positive password management code (#1097)

bama4:
Fixed fortify hard coded password in testmocking directory to suppress false positive fortify warnings about hard coded password.

### DIFF
--- a/testmocking/management/commands/add_data.py
+++ b/testmocking/management/commands/add_data.py
@@ -14,7 +14,7 @@ import re
 
 # When adding users, we expect they'll be created for test purposes. Therefore, a
 # static easy-to-remember pwd is useful.
-DEFAULT_USER_PASSWORD="GovReadyTest2019"
+DEFAULT_USER_PWD="GovReadyTest2019"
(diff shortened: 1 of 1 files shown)


gregelin:
@bama4 @peterkaminski Maybe we should move the entire 'testmocking' application out of GovReady code base. That would avoid the false positive issue.

bama4:
> @bama4 @peterkaminski Maybe we should move the entire 'testmocking' application out of GovReady code base. That would avoid the false positive issue.

My mistake, this is a false positive that should be noted, this fix is not acceptable for resolving this.

peterkaminski (commented):
This is a cool fix, but I'm concerned that an assessor might still find this section of code, and along with the base problem of "there's a password in the code", have another issue of "and why did they try to hide it by saying "pwd" instead of "password".

Closing this unmerged, and @bama4 and I think we'll just document this as a false positive.

peterkaminski:
closing unmerged, see review comment

PR #1112, opened 2020-11-19
regular users and staff need permission to view an appsource

davidpofo:
As a GovReady developer, when automating the test of GovReady UI I want to make sure regular users and staff have permission to use and view appsources assigned to them.


peterkaminski (approved):
lgtm

PR #1131, opened 2020-12-01
Cssrefactor

kerryrm:
removed inline CSS from files within the systems directory


gregelin:
Replaced by PR #1138.

Issue #1066, opened 2020-11-01
Use autocomplete to add component's control statement when authoring component control statements


gregelin:
As a GovReady user,I want to be able to be able to choose a component via autocomplete/drop-down when adding a component-control statement in the control authoring interface,so I can get the official name of the component and pre-populate the control description.

# Criteria
## Essential
* Autocomplete of components
* Being able to type-in or add a NEW component
* Populate the related statement(s) for active control from a component's related, control-specific `control_implementation_prototype` statement(s)

## Better
* Gracefully handle adding _all_ of selected component's controls to the system. This might be done by a pop-up notification telling user that all controls related to selected component have been added. May be give user option to only populate single statement(s) for active control?

## Best
* instead of just component names, additional information is available about popup
* Combine new, select existing, autocomplete, and drop down all in one widget box. For example, how AirTable adds related records.

# Screenshots
**Control authoring environment screenshot where autocomplete needed**
<img width="1105" alt="Screen Shot 2020-11-01 at 10 57 43 AM" src="https://user-images.githubusercontent.com/24979/97807928-998c0e00-1c31-11eb-86b3-c51c62e4b032.png">

**Selected components screenshot showing an implemented autocomplete**
<img width="1070" alt="Screen Shot 2020-11-01 at 11 02 28 AM" src="https://user-images.githubusercontent.com/24979/97807983-e4a62100-1c31-11eb-9855-9be4cf8934cf.png">

gregelin:
@davidpofo Note on technical implementation.

An example autocomplete for components can be found `templates/systems/components_selected.html` using the `js/select2.min.js` library which combines dropdown box with autocomplete. It would be good to try and apply the same method of autocomplete in `templates/controls/editor.html` template. 

Snippet for how the `js/select2.min.js`-based autocomplete form is rendered.
```
    <div id="tab-content" class="row row-control" style="">
      <form class="form-inline" method="post" action="/systems/{{ system.id }}/components/add_system_component">
        {% csrf_token %}
        <div id="" class="col-xs-8 col-sm-8 col-md-8 col-lg-8 col-xl-8">
        <div class="form-group">
            <label for="compoment">Add component</label>&nbsp;&nbsp;
              <select class="producer_element_id" id="producer_element_id" name="producer_element_id" style="width:340px;">
                {% for component in elements %}
                <option value="{{ component.id }}">{{ component.name }}</option>
                {% endfor %}
              </select>
        </div>
[...]

PR #1135, opened 2020-12-02
Fix OSCAL display feature flag

gregelin:
Fix enable_experimental_oscal control
    
Model method was set incorrectly requiring both enable_experimental_oscal and enable_experimental_opencontrol had to be enabled for either to show up.
    
This is now fixed.


gregelin (commented):
Small bug. Running tests locally. All looks good. There is one test failure from previous issue of routing from an invite. But this is fixed in #1066 and will be pulled in.

PR #1136, opened 2020-12-02
Json templates

gregelin:
Improvements to generating JSON output documents based on a template, including improving SSP in JSON.

### DIFF
--- a/templates/task-finished.html
+++ b/templates/task-finished.html
@@ -80,8 +80,8 @@
                     <tr>
                         <td>{{document.title}}</td>
                         <td>
-                          {% if document.format == "oscal_json" %}
-                            <a href="{{task.get_absolute_url}}/download/document/{{document.id | urlencode }}/oscal_json">OSCAL (json)</a>&nbsp;&nbsp;
+                          {% if document.format == "json" %}
+                            <a href="{{task.get_absolute_url}}/download/document/{{document.id | urlencode }}/plain">OSCAL (json)</a>&nbsp;&nbsp;
                           {% elif document.format == "oscal_xml" %}
                             <a href="{{task.get_absolute_url}}/download/document/{{document.id | urlencode }}/oscal_xml">OSCAL (xml)</a>&nbsp;&nbsp;
                           {% else %}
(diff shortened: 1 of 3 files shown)


templates/task-finished.html
```diff
@@ -80,8 +80,8 @@ <h4 class="panel-title">
                     <tr>
                         <td>{{document.title}}</td>
                         <td>
-                          {% if document.format == "oscal_json" %}
-                            <a href="{{task.get_absolute_url}}/download/document/{{document.id | urlencode }}/oscal_json">OSCAL (json)</a>&nbsp;&nbsp;
+                          {% if document.format == "json" %}
+                            <a href="{{task.get_absolute_url}}/download/document/{{document.id | urlencode }}/plain">OSCAL (json)</a>&nbsp;&nbsp;
```
peterkaminski, line 84: I'm surprised the document type here is `plain`. Shouldn't it be `oscal_json`?

(Related file = `guidedmodules/models.py`, lines 1139, 1141.)

peterkaminski (commented):
I've got a question in `templates/task-finished.html`, see comment there.

templates/task-finished.html
gregelin: @peterkaminski I need to get @JoshData help with that. The `plain` works correctly to render the JSON content to web browser. I tried for a couple hours this morning to get the JSON to download as JSON, but was not able to do so. So I want to merge the branch to produce the OSCAL version of SSP in JSON, and then work on downloading. The downloading proved more complicated than anticipate.

peterkaminski: Makes sense. I'd prefer to see `oscal_json` here, and a change to what `oscal_json` means in guidedmodules/models.py, line 1141 (where it could be made to work the same as `plain`), but I can see changing it here, too.

If we make the change here, I recommend making an issue to change it from `plain` back to `oscal_json` someday.

Merging, so we can produce the OSCAL version of SSP in JSON, and then work on downloading.

peterkaminski (approved):
Merging, see conversation at https://github.com/GovReady/govready-q/pull/1136#pullrequestreview-543147972

Issue #946, opened 2020-06-15
Same AppSource can refer to same local path of an AppSource


gregelin:
Turns out that it is possible to register a local AppSource under two different names. This is likely a problem. A path to a local app source, or a Git repo should be unique.

davidpofo:
@gregelin might be a feature in that it allows users to use the same appsource but pick and choose the compliance apps(and versions) that are used. Each new project itself will have an increment of the current number of the compliance app so if I click on the IT systems app on the left it will be 1 IT-system-1 and if I click the right it will be IT-system-2. I think there is some work here though making it clear where each compliance appsource comes from. So I suggest putting a light label next to the title like so. 
![image](https://user-images.githubusercontent.com/19982546/101029483-0f440c00-3547-11eb-965c-52c87b02841b.png)

PR #1140, opened 2020-12-03
Addressing github issue #946 where there maybe be two slugs with the …

davidpofo:
…same appsource. Removes ambiguity


davidpofo:
PR party 12/3 decided to look into the use or disuse of organizations. @gregelin help.

PR #1143, opened 2020-12-03
Remove extra bracket so button to publish project artifacts operates correctly

t-s:
An extra bracket was added to HTML so publishing artifacts was not able to be accomplished. This PR addresses that issue.

A minor change, but it allows for the button to be clicked.

Prior to change:
Screen Shot 2020-12-01 at 5 04 22 PM

After change, button works as expected.


t-s:
Any viewers - take a moment to see if the CSS should be modified w/r to the GovReady logo. I write this issue on a macbook with a smaller screen.

davidpofo (approved):
Reviewed it myself and this does solve this issue

PR #1150, opened 2020-12-03
Fixbuttons

kerryrm:
@gregelin This has the CSS to fix buttons as well as CSS fixes and inline removal of CSS from a few misc files that are not part of branding module, but needed to be altered.


davidpofo (approved):
lgtm

PR #1155, opened 2020-12-07
Docx Templating

Tom-Camp:
Templating using a `custom-reference.docx` template works. I have added an `/assets` directory and added the file in there. Ideally, we would be able to add a template to a Compliance App.

- [ ] Go to any project 
- [ ] Click on any section, for example ` Purpose of the Information Collection`
- [ ] Click on `Word (docx)`
- [ ] Open the downloaded file
- [ ] The file should contain a header
- [ ] The file should contain a footer with page numbers
- [ ] The file should contain a section for a Table of Contents


gregelin (approved):
Halleluyah!

PR #1158, opened 2020-12-07
Fix updating prototype components

davidpofo:
Get the prototype as well to update the prototype. Updating from just the prototype attribute grabs the latest implementation so the prototype is actually pointing from the last implementation.


gregelin (approved):
Code looks good. Thank you @davidpofo

PR #1165, opened 2020-12-09
Add OSCAL schema; bug fix to OSCAL generation for components

woodt:


omarabed15 (approved):

controls/views.py
```diff
@@ -285,7 +285,7 @@ def as_json(self):
             }
             # if there is a part ID, add it as a label property
             if smt.pid:
-                requirement['properties'] = dict(name='label', value=smt.pid)
+                requirement['properties'] = [dict(name='label', value=smt.pid)]
```
davidpofo, line 288: @omarabed15 looks like Tom fixed the issue here to make sure this section is an array as the nest schema expects

davidpofo (approved):
NIST schema lgtm works in omar's import code.

PR #1171, opened 2020-12-10
Added default permission to view AppSources for all users created on …

davidpofo:
…the register page


davidpofo:
need to run the sense of perms for this with greg. TBD 12/10.

gregelin (approved):
looks good

PR #1179, opened 2020-12-11
Repair errant merge that broke webdriver options for download.

woodt:


gregelin:
@woodt Not merging b/c fix was separately done with some other changes that deleted blank lines with spaces in `controls/tests.py`.

PR #1184, opened 2020-12-13
WSL has trouble using the linux paths when search for filepaths of do…

davidpofo:
…wnloaded files. Instead it uses the dos path. To fix this tempfile.tempdir needs to be set directly.


gregelin (commented):
Looks good

PR #1186, opened 2020-12-13
Create new component in component library

gregelin:
Add ElementForm for creating new components (AKA elements).
    
Add Django form, route, view, and template for creating a new component (AKA element) in the component library.

### DIFF
--- a/controls/forms.py
+++ b/controls/forms.py
@@ -59,11 +59,14 @@ class PoamForm(ModelForm):
 
 class ElementForm(ModelForm):
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.initial['element_type'] = 'system_element'
+        self.fields['element_type'].widget = forms.HiddenInput()
+
     class Meta:
         model = Element
         fields = ['name', 'full_name', 'description', 'element_type']
-        
-    element_type = forms.CharField(widget=forms.HiddenInput(), initial='system_element')
 
     def clean(self):
         """Extend clean to validate element name is not reused."""
--- a/controls/forms.py
+++ b/controls/forms.py
@@ -1,9 +1,6 @@
 from django import forms
 from django.forms import ModelForm
 from django.forms.widgets import HiddenInput
-# from django.forms import inlineformset_factory
-from django.core.exceptions import ValidationError
-from django.db.models import Exists
 
 from .models import Statement, Poam, Element
 
@@ -13,7 +10,6 @@ class StatementPoamForm(ModelForm):
         self._statement_type = kwargs.pop('statement_type', None)
         self._consumer_element = kwargs.pop('consumer_element', None)
         super().__init__(*args, **kwargs)
-        # print(self.fields['status'].__dict__)
         self.fields['status'].initial = self._status
         self.fields['statement_type'].initial = self._statement_type
         self.fields['statement_type'].widget = HiddenInput()
@@ -72,3 +68,16 @@ class ElementForm(ModelForm):
         if Element.objects.filter(name__iexact=cd['name']).exists():
             raise ValidationError("Component (aka Element) name {} not available.".format(cd['name']))
         return cd
+
+class ImportOSCALComponentForm(forms.Form):
+
+    file = forms.FileField(label="Select OSCAL file (.json)",
+        widget=forms.FileInput(
+            attrs={
+                'onchange': "fillJSONContent(this);",
+                'accept':'application/json'
+            }
+        ),
+        required=False
+    )
+    json_content = forms.CharField(label='OSCAL (JSON)', widget=forms.Textarea())
--- a/controls/forms.py
+++ b/controls/forms.py
@@ -1,6 +1,8 @@
 from django import forms
 from django.forms import ModelForm
+from django.core.exceptions import ValidationError
 from django.forms.widgets import HiddenInput
+from django.db.models import Exists
 
 from .models import Statement, Poam, Element
 
--- a/controls/forms.py
+++ b/controls/forms.py
@@ -5,7 +5,7 @@ from django.forms.widgets import HiddenInput
 from django.core.exceptions import ValidationError
 from django.db.models import Exists
 
-from .models import Statement, Poam
+from .models import Statement, Poam, Element
 
 class StatementPoamForm(ModelForm):
     def __init__(self, *args, **kwargs):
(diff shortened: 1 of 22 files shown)


gregelin:
@davidpofo @omarabed15 Could one of you review this PR, please? 

In addition to importing OSCAL, I've added a form for creating a component in the component library.

controls/views.py
```diff
@@ -5,7 +5,8 @@
 from django.conf import settings
 from django.contrib import messages
 from django.contrib.messages.views import SuccessMessageMixin
-from django.shortcuts import render
+from django.shortcuts import get_object_or_404, redirect, render
+from django.contrib.auth.decorators import login_required
```
omarabed15, line 9: This is a duplicate of line 13 below, can be removed. (This was recently merged into develop.)

controls/views.py
```diff
@@ -542,6 +544,28 @@ def system_element(request, system_id, element_id):
         }
         return render(request, "systems/element_detail_tabs.html", context)
 
+@login_required
+def new_element(request):
+    """Form to create new system element (aka component)"""
+
+    if request.method == 'POST':
+      form = ElementForm(request.POST)
+      if form.is_valid():
```
omarabed15, line 553: Might need to have an "else" statement to pair here, in case the form is invalid, to display an error message to the user and re-render the form page or redirect to the Component Library.

controls/views.py
```diff
@@ -1993,8 +2017,6 @@ def new_poam(request, system_id):
 
 def edit_poam(request, system_id, poam_id):
     """Form to create new POAM"""
-    from .forms import StatementPoamForm, PoamForm
-    from django.shortcuts import get_object_or_404
```
omarabed15, line 1997: Good catches

controls/forms.py
```diff
@@ -55,6 +55,19 @@ class Meta:
             "poam_group": "Group"
         }
 
+class ElementForm(ModelForm):
+
+    class Meta:
+        model = Element
+        fields = ['name', 'full_name', 'description', 'element_type']
```
davidpofo, line 62: For element_type it will always end up as system_element as the value going through since this is the element form. I suggest just adding a hiddeninput initial value which will provide a default for element_type. `    element_type = forms.CharField(widget=forms.HiddenInput(), initial='system_element')
`

omarabed15:
> @davidpofo @omarabed15 Could one of you review this PR, please?
> 
> In addition to importing OSCAL, I've added a form for creating a component in the component library.

@gregelin I left some comments. Overall, it looks very close, just a few minor things.

templates/components/element_form.html
```diff
@@ -0,0 +1,46 @@
+{% extends "../base.html" %}
+{% load bootstrap3 %}
+
+{% block title %}
+Welcome to Compliance Automation
```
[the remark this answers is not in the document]
gregelin: Good catch. Changing.

controls/forms.py
```diff
@@ -55,6 +55,19 @@ class Meta:
...
+
+    def clean(self):
+        """Extend clean to validate element name is not reused."""
+        cd = self.cleaned_data
+        # Validate element name does not exist case insensitive
+        if Element.objects.filter(name__iexact=cd['name']).exists():
+            raise ValidationError("Component (aka Element) name {} not available.".format(cd['name']))
```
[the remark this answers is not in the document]
gregelin: @omarabed15 Thanks for catching. I just needed to import `from django.core.exceptions import ValidationError`. The magic of Django forms automatically routes validation errors back to the user on the form (if you import the method, that is...).

controls/views.py
davidpofo: Agreed. I suggested changes to optimize imports.

controls/views.py
davidpofo: I added an else on 553 hope this covers it.

templates/systems/components_selected.html
```diff
@@ -82,7 +79,11 @@ <h2 class="control-heading">
 <script>
   // Convert select field for Add Component to jQuery Select2 box
   $(document).ready(function() {
-      $('.producer_element_id').select2();
+    $('.producer_element_id').select2();
```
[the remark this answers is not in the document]
davidpofo: Good question so in the Statements model there is a producer_element_id field so it is properly labeled here.

templates/components/element_form.html
```diff
@@ -0,0 +1,46 @@
...
+{% endblock %}
+
+{% block body-wide %}
+
+<div class="container">
+    <div class="row">
+        <div class="col-md-3 portfolio-org-col ">
+            <h2>New Component (aka Element)</h2>
```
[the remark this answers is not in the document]
gregelin: Fixed.

templates/systems/components_selected.html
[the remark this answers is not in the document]
omarabed15: I'm also not noticing this working. I'm trying out a few things, but no luck so far.

controls/forms.py
gregelin: Fixed.

controls/views.py
gregelin: The `else` block on 553 is not necessary as Django forms handles the routing back to the form page with error messages. So removing.

templates/systems/components_selected.html
```diff
@@ -82,7 +79,11 @@ <h2 class="control-heading">
...
+  });
+  // Submit the add_component form when selection changed
+  $('.producer_element_id').on("select2:select", function(e) {
```
omarabed15: This event is never being launched. If you add an "onchange" handler to the input form field in the form above, then write a function down here to call which just submits the form, that should do the trick. The field doesn't even need to be converted to a jQuery select2 field, according to my test.

templates/systems/components_selected.html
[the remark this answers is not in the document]
gregelin: @omarabed15  Class and element id are the same. So either works. Changing to element id to be more specific.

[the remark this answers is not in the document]
gregelin: @davidpofo Selecting of a component should submit the form, and reload the page with a message at top as to the result of the selection (succeed or fail). Can you please re-pull, refresh, and try again? If not working let's screenshare and validate.

omarabed15:
@gregelin @davidpofo I created this sub-PR to address the components not getting added to the Selected Components using the dropdown form. https://github.com/GovReady/govready-q/pull/1194
It's a straightforward PR, but it simplifies a few things, so I opened it for it's own discussion in case changes are requested. It should be merged before this PR is merged, if that's the desired solution.

templates/systems/components_selected.html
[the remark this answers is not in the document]
gregelin: Tested with @davidpofo and code is working and components are being added.

gregelin:
@davidpofo @omarabed15 Thanks for the careful code review and suggestions. Code is much better now.

PR #1193, opened 2020-12-14
Add default Org Defined Parameter data.

gregelin:
Add default `controls.models.OrgParams` class to support basic,
default generation of orgizational defined parameters.

Add default Organizational Defined Parameter values in
`controls/data/org_defined_parameters`.


gregelin (commented):
Looks good minor changes

PR #1198, opened 2020-12-14
Fixed bug in creating statements for component imports.

omarabed15:
Found a quick bug on creating Statements when importing an OSCAL component.


davidpofo (approved):
good catch. lgtm

PR #1197, opened 2020-12-14
All the ComponentUI tests now pass for me.

woodt:
This should be tested on MacOS and Windows again.

PR #1199, opened 2020-12-14
Control tests fix

davidpofo:


davidpofo:
> If `self.json_download` is `/tmp/foo.json`, `self.json_download.name` will be `foo.json` (a relative path) Will `os.path.isfile()` and `os.remove()` exec as intended, then?

yes the default tempdir becomes the current working directory in siteapp\tests.py line 79-81. So the download is downloading and removing from `os.getcwd()`

controls/tests.py
```diff
@@ -229,29 +228,45 @@ def test_component_download_oscal_json(self):
...
+            filetoopen = self.json_download.name
+        with open(filetoopen, 'r') as f:
             json_data = json.load(f)
             self.assertIsNotNone(json_data)
-        os.remove(self.json_download.name)
+
+        if self.json_download.is_file():
+            self.json_download.unlink()
+        elif os.path.isfile(self.json_download.name):
+            os.remove(self.json_download.name)
```
omarabed15, line 253: Is it possible to determine ahead of time which OS we're using, rather than interspersing the OS-specific checks throughout the test? I found it a little confusing to follow if these checks are done for specific test reasons, or to accommodate OS-specific functionality, and thought it might be easier to separate those two logic trees, architecturally.

siteapp/settings.py
```diff
@@ -398,6 +398,9 @@ def make_secret_key():
 	INSTALLED_APPS.append(environment["branding"])
 	TEMPLATES[0].setdefault('DIRS', [])\
 		.insert(0, os.path.join(environment["branding"], 'templates'))
+HEADLESS = False
+if environment.get("headless"):
+	HEADLESS = True
```
omarabed15, line 403: Not important, but a neat little note: In Python, you can simplify this to one line, like:
`HEADLESS = TRUE if environment.get("headless") else FALSE`

davidpofo: Works for me

controls/tests.py
davidpofo: Unfortunately no, normally I would just check the platform but within a selenium test it used the dos file path system instead of linux(WSL).

gregelin:
@davidpofo @woodt This PR and #1197 seem related. Have the two been resolved with each other?

davidpofo:
No that PR can be closed it was tom checking on my behalf if the develop at
the time could run all tests.

woodt:
I think @omarabed15 is on to something :-)  My intuition as that we could simplify all the download path handling code (and I think it would be worth doing so, since I would guess we will be writing more tests that involve upload/download.)  Maybe the three of us can discuss soon?

gregelin (approved):
Looks great! Happy to have tests fixed and fixes for CircleCI.

PR #1201, opened 2020-12-15
Fixed component OSCAL download from Component Library.

omarabed15:
Fixed the download path for downloading a component from the Component Library (i.e. when no system is involved)


gregelin (approved):
Looks good

PR #1202, opened 2020-12-15
Set default dropdown item for adding component to system

omarabed15:
Allows the first Component in the list to be selected and added to the System


gregelin (approved):
Looks good

PR #1208, opened 2020-12-17
Fix static file path handling for production

gregelin:
Create new `static-root` directory outside of `siteapp`
into which to collect static files.

Update `environment.json` in `local-example` to use new
`static_root` directory.

Remove bad path reference to select2 javascript libraries
in component library page that was causing a static file crash
in production deloyments.

Add check and raise ValidationError if `STATIC_ROOT` set to `siteapp/static`.

Uncollected static files that ship with GovReady are located in `siteapp/static`.
In development (e.g. debug = false), Django will *ignore* the STATIC_ROOT setting and
search installed application paths when resolving STATIC_URL to find actual files.

In production (e.g. debug = true), Django will use the STATIC_ROOT setting
when resolving STATIC_URL to find the path to actual files.
Also, the `manage.py collectstatic` will copy found static files into the
STATIC_ROOT path.

A duplication of files can occur in production deployments when SITE_ROOT
is defined as `siteapp/static`. Collectstatic does post-processing on files and
appends a hash and then builds a manifest for static files. As `collectstatic` command
repeatedly is run, the result is reading and aggregating static files from and into
the same directory. This will eventually cause an errors as file names grow too long
and the static file manifest and actual file names break.


siteapp/settings.py
```diff
@@ -371,13 +372,36 @@ def make_secret_key():
...
 # environment setting is present, then it's a local directory path
-# where "collectstatic" will put static files. The ManifestStaticFilesStorage
-# is activated.
+# where "collectstatic" will put static files.
+#
+# Uncollected static files that ship with GovReady are located in `siteapp/static`.
+# In development (e.g. debug = false), Django will *ignore* the STATIC_ROOT setting and
+# search installed application paths when resolving STATIC_URL to find actual files.
+#
+# In production (e.g. debug = true), Django will use the STATIC_ROOT setting
```
davidpofo, line 381: debug should never be true in production @gregelin

siteapp/settings.py
```diff
@@ -371,13 +372,36 @@ def make_secret_key():
...
+# STATIC_ROOT path.
+#
+# A duplication of files can occur in production deployments when SITE_ROOT
+# is defined as `siteapp/static`. Collectstatic does post-processing on files and 
+# appends a hash and then builds a manifest for static files. As `collectstatic` command
+# repeatedly is run, the result is reading and aggregating static files from and into
+# the same directory. This will eventually cause an errors as file names grow too long
+# and the static file manifest and actual file names break.
+#
+# The ManifestStaticFilesStorage is activated, too, for alternaye storing/serving of static assets.
```
davidpofo, line 393: alternate

PR #1218, opened 2020-12-23
OSCAL-ize control id and catalog key on import.

omarabed15:


woodt (approved):
This looks good! Out of curiosity ... there are some cases where oscalize_control_id() will return an empty string; what happens to the statement creation in that case?

As part of improving the upload UX, we may want to have a way to collect all the import errors in a report that the user could review at her leisure.  It's possible there could be a large number of them if someone tries to upload a large OSCAL component file that has issues.

omarabed15:
> This looks good! Out of curiosity ... there are some cases where oscalize_control_id() will return an empty string; what happens to the statement creation in that case?
> 
> As part of improving the upload UX, we may want to have a way to collect all the import errors in a report that the user could review at her leisure. It's possible there could be a large number of them if someone tries to upload a large OSCAL component file that has issues.

Before creating the statement, a search is done for the control ID. If it doesn't exist in the system, the statement won't get created. So, in the case where a blank "oscalized" control id is returned, no statement will get created, because the blank control id will not be found to be a valid control id.

The part about consolidating import messages is a good point. I'll work on that in a different ticket to improve the UX.

omarabed15:
Superceded by https://github.com/GovReady/govready-q/pull/1218. Closing.

PR #1228, opened 2020-12-27
Portfolio management

### LINKED ISSUE #1071: Edit and delete portfolio
As a GovReady user,
I want to be able to edit a portfolio information and delete a portfolio,
so I can better manage my workspace.

Currently, you can create a portfolio, but no editing interface to edit the portfolio name and description, or to delete the portfolio.

davidpofo:
# Pull Request

## Description

Add ability for a GovReady user to manage an existing portfolio by changing portfolio's name, description, and even delete portfolio

Closes #1071

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [X] This change requires a documentation update

## How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [ ] Unit Test - Tests moving a project to another portfolio

Scripts have been implemented to `controls/tests.py` to run it: `python manage.py test controls.tests `

## Checklist:

- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [X] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules
- [X] I have checked my code and corrected any misspellings


templates/portfolios/portfolios_table.html
```diff
@@ -28,12 +28,24 @@
     {% get_obj_perms request.user for portfolio as "perms" %}
     {% if "can_grant_portfolio_owner_permission" in perms %}
     <div class="col-md-2">Owner</div>
+        <div class="col-md-2">
+            <span class="glyphicon glyphicon-pencil"></span>&nbsp;&nbsp;&nbsp;
```
omarabed15, line 32: Should the pencil icon be inside the 'a' tag as a clickable part of the link?

templates/base-conditional.html
```diff
@@ -109,6 +109,7 @@
     {% include "bootstrap-helpers.html" %}
 
     {% include "portfolios/select-portfolio-modal.html" %}
+    {% include "portfolios/manage_portfolio_modal.html" %}
```
omarabed15, line 112: Instead of including this in a global template, perhaps this can be included only in the templates where it's used, like the portfolios_table.html template.

templates/base.html
```diff
@@ -92,6 +92,7 @@
     {% include "bootstrap-helpers.html" %}
 
     {% include "portfolios/select-portfolio-modal.html" %}
+    {% include "portfolios/manage_portfolio_modal.html" %}
```
omarabed15, line 95: Same comment as previous

siteapp/views.py
```diff
@@ -1418,6 +1418,59 @@ def new_portfolio(request):
...
+@login_required
+def edit_portfolio(request, pk):
+    """Form to create new portfolios"""
+
+    if request.method == 'GET':
+        portfolio = Portfolio.objects.get(pk=pk)
+        form = PortfolioForm(request.POST or None, instance=portfolio, initial={'portfolio': portfolio.id})
+        project_form = ProjectForm(request.user, initial={'portfolio': portfolio.id})
+        if form.is_valid():
+            form.save()
```
omarabed15, line 1447: Since this is still checking for validity, it makes me think the "clean" function should still be present in the form.

templates/portfolios/detail.html
```diff
@@ -28,15 +28,19 @@
...
-      <a id="new-project" onclick="show_select_portfolio_modal(); return false;" href="#"" class=" btn
+        <a id="new-project" onclick="show_select_portfolio_modal(); return false;" href="#" class=" btn
         btn-success">Start a project</a>
-      {% if can_invite_to_portfolio %}
-      <a href="#" onclick="invite_user_into_portfolio(); return false;" id="grant-portfolio-access"
-        class="btn btn-success"><i class="glyphicon glyphicon-plus"></i> Grant access to member</a>
-      {% endif %}
+        {% if can_change_portfolio %}
+            <a id="manage-portfolio" onclick="show_manage_portfolio_modal('{{portfolio.id}}'); return false;" href="#" class=" btn
+            btn-success">Manage Portfolio</a>
```
omarabed15, line 38: This is definitely a nit-pick, but it seems odd to split the line here. If the line is too long, maybe split at the start of the "class" attribute so it's clear what all the classes here. But again, very small nitpick.

templates/portfolios/edit_form.html
```diff
@@ -0,0 +1,50 @@
...
+    </div>
+</div>
+
+
+{{ block.super }}
+{% endblock %}
+
+{% block scripts %}
+<script>
+    $('#id_name').on('keyup', function () {
```
omarabed15, line 40: Comments here would help as a quick snapshot of what this is doing.

omarabed15: Also, is it necessary to run this code on every key press on this element? It's making everything lowercase on each keypress, but it might be easier to add an on-click event to the "submit" button of the form ("Update") and do this "to lowercase" process once at the end.

templates/portfolios/manage_portfolio_modal.html
```diff
@@ -0,0 +1,35 @@
...
+      </div>
+      <form role="form" method="POST">
+          {% csrf_token %}
+        <div class="modal-body">
+          <label for="id_portfolio">{{ project_form.portfolio.label }}</label>:&nbsp;
+          {{ project_form.portfolio }}
+         </div>
+         <div class="modal-footer">
+           <a id="edit_portfolio_submit" class="btn btn-success btn-submit" type="submit">Edit</a>
+           <a id="delete_portfolio_submit"  class="btn btn-danger btn-submit confirm-delete" type="submit" onclick="return confirm('Are you sure you want to delete this?')">Delete Portfolio</a>
```
omarabed15, line 16: I like this method for giving a confirmation message. I had done something much more complicated for one of my tickets, but this is nice and clean.

templates/portfolios/portfolios_table.html
omarabed15: If not, it seems like it will get crowded to have multiple unclickable pencil icons in the table. Maybe just have one as a separation from the readable and editable actions on the table, to avoid crowding? Just a thought.

templates/portfolios/portfolios_table.html
```diff
@@ -28,12 +28,24 @@
...
+            <a id="new-project" class="portfolio-project-link" onclick="show_manage_portfolio_modal('{{portfolio.id}}'); return false;" href="#">Manage Portfolio</a>
+        </div>
     {% elif "change_portfolio" in perms%}
     <div class="col-md-2">Portfolio Member</div>
+    <div class="col-md-2">
+        <div class="col-md-2">
+            <span class="glyphicon glyphicon-pencil"></span>&nbsp;&nbsp;&nbsp;
```
omarabed15, line 39: I'm not a big fan of the use of &nbsp; to add space, but I know we're doing that throughout this project. I think that is getting swapped out though, to be replaced by CSS styling code, so if this could be replaced by spacing in CSS, that'd be great.
Same comment throughout, in a few spots.

siteapp/forms.py
```diff
@@ -27,13 +27,6 @@ class Meta:
         model = Portfolio
         fields = ['title', 'description']
 
-    def clean(self):
-        """Extend clean to validate portfolio name is not reused."""
```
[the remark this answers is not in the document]
davidpofo: True, the issue I had was trying to update the description while keeping the title the same. This validation prevents the update from happening. Is there a way to avoid using clean for this particular use-case?

templates/portfolios/portfolios_table.html
davidpofo: So this is an attempt to be consistent with how a user chooses a portfolio to view. They click on the name and not the icon or combination of name & icon. The reason the icon is there is to avoid the user not recognizing that that section is where you can pick a portfolio to edit/delete.

siteapp/forms.py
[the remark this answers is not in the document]
omarabed15: I would think it should continue to block that case, so that people don't rename a Portfolio to have the same name as another. Maybe all it needs is to keep this validation and add a message when the validation fails, such as "Name already used. Cannot save Portfolio with this name. Please try another." and it could redirect back to the Portfolio edit page.

templates/portfolios/portfolios_table.html
omarabed15: Oh okay, I didn't notice the folder icon for viewing the portfolio. In that case, this makes sense to me. I might still suggest a single pencil icon before all the portfolio edit commands, rather than one before each command, to avoid crowding, but it makes sense to me now why it's not clickable, for consistency. Thanks!

siteapp/forms.py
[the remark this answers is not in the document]
davidpofo: Adding a updateportfolio hidden input to check for in the clean method

siteapp/views.py
```diff
@@ -1418,6 +1418,59 @@ def new_portfolio(request):
...
+@login_required
+def delete_portfolio(request, pk):
+    """Form to delete portfolios"""
+    if request.method == 'GET':
+        portfolio = Portfolio.objects.get(pk=pk)
+        Portfolio.objects.get(pk=pk).delete()
+
+        logger.info(
+            event="delete_portfolio",
+            object={"object": "portfolio", "id": portfolio.id, "title": portfolio.title},
```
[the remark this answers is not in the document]
omarabed15: Another minor nit-pick: This is technically logging before the delete happens. If there is an error with the delete (such as the portfolio not being found) the log will still happen, which might be confusing. I don't think it's likely, since the portfolio is retrieved just 1 step prior, but something to consider. I was initially thinking to store the portfolio _properties_ in variables before the deletion, but not a huge deal.

siteapp/forms.py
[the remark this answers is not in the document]
omarabed15: But the same "clean" validation of the name should happen on both create and edit, correct? This new logic seems to skip this validation if the portfolio name is being edited. I think we still need to check for name duplicates on edit, and just inform the user if the name is invalid (on both edit and create).

templates/base-conditional.html
omarabed15: I see now that this is consistent with the "select portfolio" modal. I'm not a fan of that logic in general, but maybe this should all be refactored later, if not done now.

templates/portfolios/portfolios_table.html
davidpofo: Not seeing an easy way to do that since we have 3 different member types Owner, portfolio member, project member

templates/base-conditional.html
kerryrm: I concur.

templates/portfolios/portfolios_table.html
omarabed15: The last comment may have been intended somewhere else. Here, I'm just mentioning adding styling code to the icon instead of the "&nbsp;"

omarabed15: The last comment may have been intended under a different issue, but this issue seems to have been resolved, so I'm resolving this one.

templates/portfolios/edit_form.html
omarabed15: Looks like this was removed. I'm a fan of that. Resolving this issue.

siteapp/views.py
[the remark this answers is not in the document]
davidpofo: I moved the logging up so this should be resolved

siteapp/views.py
```diff
@@ -1393,31 +1393,86 @@ def portfolio_list(request):
...
+def delete_portfolio(request, pk):
+    """Form to delete portfolios"""
+    if request.method == 'GET':
+        portfolio = Portfolio.objects.get(pk=pk)
+        logger.info(
+            event="delete_portfolio",
+            object={"object": "portfolio", "id": portfolio.id, "title": portfolio.title},
+            user={"id": request.user.id, "username": request.user.username}
+        )
+        Portfolio.objects.get(pk=pk).delete()
```
omarabed15, line 1433: Now the `Portfolio.objects.get(pk=pk)` query is being run twice, rather than capitalizing on the first one run. I'd suggest something like this:

`portfolio = Portfolio.objects.get(pk=pk)
  pid = portfolio.id
  ptitle = portfolio.title
  portfolio.delete()
  logger.info(
    event="delete_portfolio",
    object={"object": "portfolio", "id": pid, "title": ptitle},
    user={"id": request.user.id, "username": request.user.username}
  )`

This way, you're able to:
- Only run the 'get' query once
- Only preserve the data you need from the portfolio before it is deleted (rather than reusing a "deleted" portfolio variable after it's deleted)
- Only log the deletion after it has actually happened (in case there's an error and the delete doesn't happen, for some reason).

siteapp/views.py
```diff
@@ -1393,31 +1393,86 @@ def portfolio_list(request):
...
+            form.save()
+
+            logger.info(
+                event="edit_portfolio",
+                object={"object": "portfolio", "id": portfolio.id, "title": portfolio.title},
+                user={"id": request.user.id, "username": request.user.username}
+            )
+            portfolio.assign_owner_permissions(request.user)
+            logger.info(
+                event="new_portfolio assign_owner_permissions",
```
omarabed15, line 1454: Is "new_portfolio" needed here? Since it's just editing the portfolio, should that part be removed to just say "assign_owner_permissions"? Especially since an "edit_portfolio" event is logged just 2 lines before.

templates/base-conditional.html
omarabed15: @davidpofo Even if the entire issue isn't resolved here, this PR would be a good place to move these specific new include statements into somewhere like the portfolios_table.html template, or wherever else they're used.

omarabed15 (commented):
So close! It's looking good.

siteapp/views.py
davidpofo: Yup permissions can be assigned through other functions

PR #1230, opened 2020-12-28
incorporating django-import-export for project import/export. Still n…

davidpofo:
…eeds validation( also note it can't import elements/components info).


templates/controls/import_modal
```diff
@@ -0,0 +1,34 @@
...
+<script>
+  function show_project_modal(id,callback) {
+    var m = $('#project_modal');
+    m.modal();
+    {#$("#edit_portfolio_submit").attr("href", "/portfolios/"+id+"/edit");#}
+    {#$("#delete_portfolio_submit").attr("href", "/portfolios/"+id+"/delete");#}
+  }
+
+</script>
+{% endblock %}
```
Yodart, line 34: File type should be `.html`

templates/project.html
```diff
@@ -479,7 +485,7 @@ <h3 class="question-group-title">{{group.title}}</h3>
 {% endblock %}
 
 {% block modals %}
-
+{% include "controls/import_modal" %}
```
Yodart, line 488: `{% include "controls/import_modal.html" %}`

templates/controls/import_modal
```diff
@@ -0,0 +1,34 @@
...
+    <button class="btn btn-primary" type="submit">Import</button>
+  </form>
+      </div>
+    </div>
+  </div>
+</div>
+
+{% block scripts %}
+<script>
+  function show_project_modal(id,callback) {
```
Yodart, line 26: Lines 29 and 30 are never used, those elements do not exist on this file.

Yodart (commented):
Minor alterations needed. Outside of the changes commented on each file I believe `django-import-export` should be added to the project's `requirements.txt`

templates/controls/import_modal.html
```diff
@@ -5,7 +5,7 @@
         <button type="button" class="close" data-dismiss="modal" aria-hidden="true">&times;</button>
         <h4 class="modal-title" id="project_modal_title">Select a Project</h4>
       </div>
-<form method="post" enctype="multipart/form-data" action="/systems/{{project.system.id}}/import">
+<form method="post" enctype="multipart/form-data" action="/systems/{{project.system.id}}/im ">
```
Yodart, line 8: `controls/urls.py` expects `/import` not `/im`

Yodart (changes requested):

Yodart (approved):

controls/views.py
```diff
@@ -2400,3 +2402,114 @@ def poam_export(request, system_id, format='xlsx'):
...
+            # Retrieve identified System
+            system = System.objects.get(id=system_id)
+            #Import questionnaire data
+            log_output = []
+            try:
+                from collections import OrderedDict
+                data = json.loads(imported_jsondata, object_pairs_hook=OrderedDict)
+            except Exception as e:
+                log_output.append("There was an error reading the export file.")
+            else:
```
omarabed15, line 2463: Interesting. I hadn't seen a try-except-else before. I learned something new!

controls/views.py
```diff
@@ -2400,3 +2402,114 @@ def poam_export(request, system_id, format='xlsx'):
...
+            # Log output
+            logger.info(
+                event="project JSON import",
+                object={"object": "project", "id": project.id, "title": project.title, "log_output": log_output},
+                user={"id": request.user.id, "username": request.user.username}
+            )
+
+            # Load and get the components then dump
+            for k, val in enumerate(json.loads(imported_jsondata).get('component-definitions')):
+                oscal_component_json = json.dumps(json.loads(imported_jsondata).get('component-definitions')[k])
```
omarabed15, line 2479: This is doing a repeated `json.loads(imported_jsondata)` call. Probably best to store that in a variable outside of the 'for' loop, then use that variable throughout.

controls/views.py
```diff
@@ -2400,3 +2402,114 @@ def poam_export(request, system_id, format='xlsx'):
...
+                result = ComponentImporter().import_component_as_json(oscal_component_json, request)
+
+        return HttpResponseRedirect("/systems/{}/controls/selected".format(system_id))
```
omarabed15, line 2482: Note: You can also use an "f-string" in Python3, which I find a bit easier to read. Although, I'm not sure which is "better". But here's how that might look:
`return HttpResponseRedirect(f"/systems/{system_id}/controls/selected")`

controls/views.py
```diff
@@ -2400,3 +2402,114 @@ def poam_export(request, system_id, format='xlsx'):
     else:
         # User does not have permission to this system
         raise Http404
+
+def project_import(request, project_id):
+    """
+    Import an entire project's components and control content
+    """
+    project = Project.objects.get(id=project_id)
```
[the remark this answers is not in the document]
omarabed15: I guess I'm confused why a project_id is even required for this function, and where it gets passed in from. I can see that the template passes it in, but how would that template have access to the project id before the project has been imported? I assume that project id can be parsed out from the json, so it probably has to be done after the json is loaded in this function, but not before.

templates/controls/import_modal.html
```diff
@@ -0,0 +1,34 @@
+<div id="project_modal" class="modal" tabindex="-1" role="dialog" aria-labelledby="project_modal_title" aria-hidden="true">
+  <div class="modal-dialog">
+    <div class="modal-content">
+      <div class="modal-header">
+        <button type="button" class="close" data-dismiss="modal" aria-hidden="true">&times;</button>
+        <h4 class="modal-title" id="project_modal_title">Select a Project</h4>
+      </div>
+<form method="post" enctype="multipart/form-data" action="{% url 'project_import' project.id %}">
```
omarabed15, line 8: How does the import project modal have access to a project id before the project is imported? What project is this id coming from?

controls/views.py
```diff
@@ -2400,3 +2402,114 @@ def poam_export(request, system_id, format='xlsx'):
...
+            component = OSCALComponentSerializer(element, impl_smts).as_json()
+            oscal_comps.append(component)
+
+    # TODO: multiple export types
+
+    questionnaire_data = json.dumps(project.export_json(include_metadata=True, include_file_content=True))
+    data = json.loads(questionnaire_data)
+    for oscal_comp in oscal_comps:
+        data['component-definitions'] = json.loads(oscal_comp)
+    data['component-definitions'] = [json.loads(oscal_comp) for oscal_comp in oscal_comps]
```
omarabed15, line 2511: These seem like duplicates of each other. Are both really needed?

siteapp/urls.py
```diff
@@ -48,8 +48,8 @@
     url(r'^projects/(\d+)/__rename$', views.rename_project, name="rename_project"),
     url(r'^projects/(\d+)/__delete$', views.delete_project, name="delete_project"),
     url(r'^projects/(\d+)/__admins$', views.make_revoke_project_admin, name="make_revoke_project_admin"),
-    url(r'^projects/(\d+)/__export$', views.export_project, name="export_project"),
-    url(r'^projects/(\d+)/__import$', views.import_project_data, name="import_project_data"),
+    url(r'^projects/(\d+)/__export$', views.export_project_questionnaire, name="export_project_questionnaire"),
+    url(r'^projects/(\d+)/__import$', views.import_project_questionnaire, name="import_project_questionnaire"),
```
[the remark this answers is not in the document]
davidpofo: Yeah they are necessary to maintain huge functionality differences between export and import

templates/controls/import_modal.html
davidpofo: Right so in the modal we allow the user to choose to make a new project (with the checkbox) or update the current one. The latter is allowed with this project.id and if they want to create a new project we need the system and portfolio information from the current project we are on.

controls/views.py
davidpofo: true got right of the first one.

controls/forms.py
```diff
@@ -88,3 +88,32 @@ class ImportOSCALComponentForm(forms.Form):
...
+    file = forms.FileField(label="Select project file (.json)",
+        widget=forms.FileInput(
+            attrs={
+                'onchange': "fillJSONContent(this);",
+                'accept':'application/json'
+            }
+        ),
+        required=False
+    )
+    json_content = forms.CharField(label='Project (JSON)', widget=forms.Textarea())
```
omarabed15, line 103: Might as well benefit off the other forms that do the same thing! I like it. It makes me wonder if we should have a base "ImportForm" class, and then extend from that to make our separate "Component" and "Project" importers. But that's out of scope of this task. It's probably a TODO down the line. Just a thought.

controls/forms.py
```diff
@@ -88,3 +88,32 @@ class ImportOSCALComponentForm(forms.Form):
...
+        cleaned_data = super(ImportProjectForm, self).clean()
+        file = cleaned_data.get('file')
+
+        if file:
+            filename = file.name
+            if filename.endswith('.json'):
+                print('File is a .json file')
+            else:
+                print('File is NOT a .json file')
+                raise forms.ValidationError("File is not a mp3. Please upload only mp3 files")
```
omarabed15, line 117: Is it necessary to overwrite this "clean" method and have these print statements? It seems excessive to me, since the formField already does validation of the file type with this attribute: `'accept':'application/json'`

controls/views.py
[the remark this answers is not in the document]
omarabed15: @davidpofo Thanks for answering this. I think long-term this may benefit from being refactored to not require a project_id (in case we allow project imports from outside of a project/system), but for now this works.

controls/views.py
```diff
@@ -2400,3 +2402,110 @@ def poam_export(request, system_id, format='xlsx'):
...
+
+def project_import(request, project_id):
+    """
+    Import an entire project's components and control content
+    """
+    project = Project.objects.get(id=project_id)
+    # Retrieve identified System
+    if request.method == 'POST':
+        project_data = request.POST['json_content']
+        file_format = "JSON"# TODO: Check the file format properly
```
omarabed15, line 2414: This probably isn't needed if the form validates the file type. Just a thought.

controls/views.py
```diff
@@ -2400,3 +2402,114 @@ def poam_export(request, system_id, format='xlsx'):
...
+    # Retrieve identified System
+    if request.method == 'POST':
+
+        project_data = request.FILES['import_project_data']
+        file_format = request.POST['file-format']
+        importcheck = False
+        if "importcheck" in request.POST:
+            importcheck = request.POST["importcheck"]
+        if file_format == 'CSV':
+            pass
```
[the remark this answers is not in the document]
omarabed15: The way it is now, the logic might confuse some people (if we don't get back to this code for a while), because it's hardcoded as "JSON" above and then there's a check here for "CSV" that doesn't do anything. I think for now, a TODO is most clear

siteapp/urls.py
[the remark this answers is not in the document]
omarabed15: But I'm not seeing any calls to the `.../__export` or `.../__imports` urls, so are these even being used?

templates/controls/import_project_modal.html
```diff
@@ -0,0 +1,53 @@
...
+                    <label for="{{ import_project_form.json_content.id_auto }}">{{ import_project_form.json_content.label }}:</label>
+                    <br>
+                    {{ import_project_form.json_content }}
+
+                    <div class="modal-footer">
+                        <input id="import_component_submit" class="btn btn-success btn-submit" type="submit" value="Import"/>
+                        <button type="button" class="btn btn-default" data-dismiss="modal">Cancel</button>
+                        {{ import_project_form.importcheck }}
+                        {{ import_project_form.importcheck.label }}
+                        <img id="import_loading_spinner" src="{% static "img/icons/spinner.gif" %}"/>
```
omarabed15, line 23: Nice! Might as well reuse what we have. This makes me think we should have a base "importer_form" template, and extend it based on which form we want to use (i.e. Component importer or Project importer), but that's probably out of scope for now. It's a refactoring TODO to keep in mind.

templates/components/import-component-modal.html
```diff
@@ -32,17 +33,7 @@ <h4 class="modal-title" id="invitation_modal_title">Import OSCAL Component</h4>
...
-    var reader = new FileReader();
-    reader.readAsText(filecontents);
-    reader.onload = function(e) {
-        $('#{{import_form.json_content.auto_id}}').val(e.target.result);
-    };
-  }
 
-  function show_spinner() {
-    $("#import_loading_spinner").show();
-  }
```
omarabed15, line 46: These are still needed by the Component Importer on the Component Library, so they can't be deleted.

omarabed15: If it's a matter of not having these duplicated across both the component and project import forms, I think this material should be separated into an extendable base template, because we can't always ensure that both the component importer and project importer will be available on the same page, so we need to have this JS accessible where it's needed.

templates/project-base.html
```diff
@@ -133,6 +133,30 @@ <h1 id="project-title">
...
+                $('#{{import_project_form.json_content.auto_id}}').val(e.target.result);
+            };
+        }
+
+        function show_spinner() {
+            $("#import_loading_spinner").show();
+        }
+    </script>
+{% endblock %}
+{% include "controls/import_project_modal.html" %}
```
omarabed15, line 158: This code is being duplicated between this file and the import_project_modal (which is included in the last line here), and ideally also with the component import template (where it shouldn't be removed), so that's looking like a lot of JS duplication. This makes me think the JS should be separated into a template somehow. Or maybe these entire modals should be separated into their own dedicated "import_modal" template.

omarabed15 (changes requested):

controls/views.py
```diff
@@ -2400,3 +2402,103 @@ def poam_export(request, system_id, format='xlsx'):
...
+            component = OSCALComponentSerializer(element, impl_smts).as_json()
+            oscal_comps.append(component)
+
+    # TODO: multiple export types
+
+    questionnaire_data = json.dumps(project.export_json(include_metadata=True, include_file_content=True))
+    data = json.loads(questionnaire_data)
+    data['component-definitions'] = [json.loads(oscal_comp) for oscal_comp in oscal_comps]
+    response = JsonResponse(data, json_dumps_params={"indent": 2})
+    filename = project.title.replace(" ", "_") + "-" + datetime.now().strftime("%Y-%m-%d-%H-%M")
```
peterkaminski, line 2502: Regarding `project.title.replace(" ", "_")`:

There are many ways filenames can have bad characters or substrings.

So instead of the simple `replace`, consider using [pathvalidate](https://pypi.org/project/pathvalidate/) or something similar.

controls/forms.py
```diff
@@ -88,3 +89,25 @@ class ImportOSCALComponentForm(forms.Form):
...
+            }
+        ),
+        required=False
+    )
+    json_content = forms.CharField(label='Project (JSON)', widget=forms.Textarea(), help_text="The JSON necessary for importing a project.")
+    importcheck =  forms.BooleanField(label="Import as a new project", required=False, help_text="If checked the current import will become a new project.")
+    #appsource_slug = forms.CharField(label="Enter the slug name of App Source", help_text="Needs a unique slug name for the App Source.")
+    #appsource_type = forms.CharField(label="Enter the type of App Source", help_text="Enter the App Source type.")
+   # appsource_path = forms.FilePathField(label="Choose an App Source path", path='.', recursive=True, allow_files=False, allow_folders=True, help_text="Select the App Source Associated with this project.")
+   # appsource_compapp = forms.FilePathField(label="Choose the compliance app from your App Source", path='.', recursive=True, allow_files=False, allow_folders=True, help_text="Select the App Source compliance app.")
```
omarabed15, line 109: Should remove these, for cleanliness

controls/views.py
```diff
@@ -2457,7 +2457,16 @@ def project_import(request, project_id):
...
+            project_names = Element.objects.filter(element_type="system").values_list('name', flat=True)
+            # If it is a new title just make that the new system name otherwise increment
+            new_title = new_project.title
+            if new_title not in project_names:
+                new_title = new_project.title
+            else:
+                while new_title in project_names:
+                    new_title = increment_element_name(new_title)
+
+            element.name = new_title
```
omarabed15, line 2469: Nice, I like it. Good reuse and broadening of the increment names function

requirements.txt
```diff
@@ -67,83 +67,63 @@ cffi==1.14.4 \
     --hash=sha256:ec80dc47f54e6e9a78181ce05feb71a0353854cc26999db963695f950b5fb375 \
     --hash=sha256:f032b34669220030f905152045dfa27741ce1a6db3324a5bc0b96b6c7420c87b \
     --hash=sha256:f60567825f791c6f8a592f3c6e3bd93dd2934e3f9dac189308426bd76b00ef3b \
-    --hash=sha256:f803eaa94c2fcda012c047e62bc7a51b0bdabda1cad7a92a522694ea2d76e49f
+    --hash=sha256:f803eaa94c2fcda012c047e62bc7a51b0bdabda1cad7a92a522694ea2d76e49f \
     # via cryptography
-chardet==4.0.0 \
-    --hash=sha256:0d6f53a15db4120f2b08c94f11e7d93d2c911ee118b6b30a04ec3ee8310179fa \
-    --hash=sha256:f864054d66fd9118f2e67044ac8981a54775ec5b67aed0441892edb553d21da5
+chardet==3.0.4 \
```
omarabed15: Are changes to requirements.txt intended to be captured in this PR? Do these dependency changes relate to the work being done here? I only ask because we've had dependency issues in the past, so it might be good to isolate them if they're not needed here.

siteapp/urls.py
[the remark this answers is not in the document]
davidpofo: yes in the project's settings.

requirements_mysql.txt
```diff
@@ -1,7 +1,7 @@
...
-    --hash=sha256:4861e33b9fad83dadf38feae77a14106365fc8c354ec4300fab5d8ccfbee4b4e \
-    --hash=sha256:4d91dde55230a512937aafc5981c6c11c0efad2c6b3d33e83287208c8bdab367 \
-    --hash=sha256:8944d9008c29abfa43e3c97551416ea23b2ccb017af85682b79f34fc26dc43ae \
-    --hash=sha256:8df057b08fc27d8f7106bfa997d0a21e2acef017f905f06d6fb0aa6a20d4d2b2 \
+mysqlclient==2.0.3 \
+    --hash=sha256:0ac0dd759c4ca02c35a9fedc24bc982cf75171651e8187c2495ec957a87dfff7 \
+    --hash=sha256:3381ca1a4f37ff1155fcfde20836b46416d66531add8843f6aa6d968982731c3 \
+    --hash=sha256:71c4b330cf2313bbda0307fc858cc9055e64493ba9bf28454d25cf8b3ee8d7f5 \
+    --hash=sha256:f6ebea7c008f155baeefe16c56cd3ee6239f7a5a9ae42396c2f1860f08a7c432 \
+    --hash=sha256:fc575093cf81b6605bed84653e48b277318b880dc9becf42dd47fa11ffd3e2b6
```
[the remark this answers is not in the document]
davidpofo: @gregelin says it's okay for this instance

templates/project-base.html
davidpofo: Ended up I needed to send an import_name so I made some adjustments with the js function fillProjectJSONContent

controls/views.py
davidpofo: quote(filename) should take care of any problematic filenames.

controls/tests.py
```diff
@@ -895,3 +895,107 @@ def create_simple_import_record(self):
...
+
+        # Project title to search for in file names
+        project_title = "I_want_to_answer_some_questions_on_Q._3"
+        file_system = os.listdir(os.getcwd())
+        # Assert there is a file
+        for file_name in file_system:
+            if file_name ==project_title:
+                project_title = file_name
+
+        self.assertIn(file_name, file_system)
```
omarabed15, line 1001: You could also take a look at how the `test_component_download_oscal_json` test works (in the same controls/tests.py file) to check whether or not a specific file was downloaded, rather than looping through an entire directory, just in case that directory gets full and that loop takes a while

controls/tests.py
```diff
@@ -895,3 +895,107 @@ def create_simple_import_record(self):
...
+        self._login()
+        self._new_project()
+
+        # Checks the number of projects and components before the import
+        self.assertEqual(Project.objects.all().count(), 3)
+        self.assertEqual(Element.objects.all().exclude(element_type='system').count(), 0)
+
+        ## Update current project
+        # click import project button, opening the modal
+        self.click_element("#action-buttons\ action-row > div.btn-group > button:nth-child(4)")
```
omarabed15, line 955: My concern here is, what if the order of the buttons change, or a new button is added? Since this is looking for the 4th button, it might break the test. Maybe look for the button by the text "Import Project"?

controls/tests.py
```diff
@@ -895,3 +895,107 @@ def create_simple_import_record(self):
...
+        """
+        Testing the update of a project through project JSON import and ingestion of components and their statements
+        """
+
+        # login as the first user and create a new project
+        self._login()
+        self._new_project()
+
+        # Checks the number of projects and components before the import
+        self.assertEqual(Project.objects.all().count(), 3)
```
omarabed15, line 950: My concern here is if other tests add (or delete) more projects in the future, then this "3" won't be correct and the test will (wrongly) break. We might have to consider storing the "pre-import" value in a variable, and then doing an assert for the difference afterwards, like:
`project_count_before_import = Project.objects.all().count()`
`# Do import stuff`
`self.assertEqual(Project.objects.all().count(), project_count_before_import + 1)`

That would be the case with Elements and Statements. It's happening throughout the tests, so that's probably some technical debt we'll have to resolve down the road throughout the project.

PR #1231, opened 2020-12-28
Track imports

omarabed15:
Creates a new ImportRecord model, for tracking imports and deleting them. Imports can be managed from the "Import Component" modal dialog in the Component Library. We can add this link to more convenient places later on.


gregelin (approved):
Looks good!

PR #1234, opened 2020-12-29
Add an autocomplete in component library to look up controls across multiple catalogs for writing a control implementation statement.

gregelin:
Support a Select2 autocomplete dropdown selection box in
the component library to assign a control when authoring
a new component control implementation statement for a component
in the library.

In the component library, the control selection drop down must
support controls from multiple catalogs because we want to be
able to associate a component (e.g. system element) with any
control catalog.

Add 'label' value to `oscal.Catalog.cx.get_flattened_controls_all_as_dict`.

Added route `controls/api/controlsselect/` and view `api_controls_select`
to get list of controls.

Modified view `save_smt` to just save prototype statement when
statement is being created in the component library.

Modified template `templates/components/element_detail_tabs.html`
to use jQuery select2 for autocomplete and search of catalog of
controls to add a control to a component.

Update hidden sid_class field with catalog human readable name.
Add hidden field `form_source` to identufy to save smt view that
we are receiving form submission from component library.

Still To Do:
- Faster querying of controls for Autocomplete
- case independent search


controls/views.py
```diff
@@ -1269,53 +1303,60 @@ def save_smt(request):
...
+            # Serialize saved data object(s) to send back to update web page
+            # The submitted form needs to be updated with the object primary keys (ids)
+            # in order that future saves will be treated as updates.
+            from django.core import serializers
+            serialized_obj = serializers.serialize('json', [statement, ])
 
     # Return successful save result to web page's Ajax request
     return JsonResponse(
-        {"status": "success", "message": statement_msg + " " + producer_element_msg + " " + statement_element_msg,
+        {"status": "success", "message": statement_msg + " " + producer_element_msg + " " + statement_element_msg + statement_del_msg,
```
bama4, line 1359: Could do `"  ".join([statement_msg, producer_element_msg, statement_element_msg, statement_del_msg])` instead

davidpofo:
PR Party 1/7 also need to filter out implementations by catalog. case-independent search.

Issue #1237, opened 2020-12-30
Slowdown on project page (single project), with many projects/users


HeatherSoron:
**Describe the bug as a Story**
As a user of GovReady, I want to be able to use the software even with a large amount of data. With a lot of data in the system, it currently takes too long to load a single project's page (i.e., to render the `project` view in siteapp/views.py). I've observed multi-second load times.

**To Reproduce**
1. Set up a large amount of data, using `python3 manage.py add_data` to add (e.g.) 2000 users plus 400 projects (note: takes around 50 minutes for me, largely because of this issue; reproduction is likely feasible with a smaller project count).
2. Load the page for a single project, OR do anything else to invoke the `project` view.
3. Observe that the time to execute and render the view is far too high; easily multiple seconds.

**Expected behavior**
Time taken to load the project page stays reasonable, until a much-higher number of projects/users.

## Developer hints

There's a loop in the `has_upgrade_priv` method of AppVersion, which iterates over all projects linked to the AppVersion in question and executes a database query (via a function call) for each project. Unencapsulating that, and converting to a single query, would likely result in a dramatic speedup. Or equivalently, providing a static method on the Project model which operates in batch mode rather than iterating within Python.

I also noticed a lot of time taken in the `render` method, called by the view itself, but this was less significant than `has_upgrade_priv`. The slowness in `render` may also have been a profiler artifact, rather than an actual issue.

I also noticed that profiling the code using pyinstrument degraded performance even further, *but* it was still a useful way to identify where the bottleneck was occurring.

## Acceptance criteria 

GIVEN an installation of GovReady-Q, with a large amount of data added via `python3 manage.py add_data` (e.g., 2000 users and 400 projects)...
WHEN I log in as the admin user and navigate to a single project's page (which is the one rendered by the `project` view in siteapp/views.py)...
THEN page load time is significantly sped up from the current state. (under 1 second?)

**Additional context**
The problem was originally identified while working on automated data generation, hence the mention of a view rather than a URL.

PR #1236, opened 2020-12-30
Fix and optimize data generation, partially addresses #1137

HeatherSoron:
This branch updates data generation to support adding 2000 users and 400 projects, with "under 1 hour" as an informal goal (further speedup may be done in other pull requests).

Using the code on this branch, use `python3 manage.py add_data` to create the desired number of users and projects. I recommend that one uses 0 delay between actions, for large-scale data generation.

The main changes included are:
* Renames the `datamocking` module to `loadtesting`
* Fixes some parts of data generation that had become non-functional
* Optimizes certain parts of data generation code (only), to work around slowdowns with a large amount of data
* Incorporates the `pyinstrument_add` branch, because pyinstrument was important for the sake of identifying bottlenecks

I do see a merge conflict, but it looks like it's in `requirements.in` (and `requirements.txt`), so I'll defer to the person merging this to resolve that - unless it's desired that I resolve it prior to review, of course.

PR #1239, opened 2020-12-30
Speed up `has_upgrade_priv` (helps with #1237)

HeatherSoron:
Here's a proposed improvement for #1237. For me it's yielding a speedup (for just this method) from 3.4s down to 0.5s (or with pyinstrument's overhead, from 8.3s down to 1.8s), when I have 412 projects that were created via automated data generation. Times measured in pyinstrument, so they may be a bit inflated compared to real-world times.

~~I suspect there's still more room for improvement~~ [correction: when *not* using pyinstrument, this might be okay, actually], but this was fairly quick to put together, and should help significantly with data generation times as well as with some sources of site slowness.

I've spot-tested to verify that functionality is unchanged (only execution speed), but that aspect could probably do with better testing.

I also tested with only 3 projects on a fresh DB, and verified that speed is good in that situation as well. It does look like it might be faster there *too* (not too surprising), although it's not a performance bottleneck with such a small number of projects.

Includes PR #1236, but I could rebase this onto a different branch if that's preferred (there's no actual dependence on the code from that branch).


gregelin:
@HeatherSoron I merged #1236 int develop. Can you please merge most recent develop bak into this PR and resolve conflicts?

HeatherSoron:
@gregelin I've merged current develop into this PR, should be clean now.

davidpofo:
@peterkaminski  is looking into if it is faster PR party 1/14

gregelin (changes requested):
@HeatherSoron I am bit confused about `BeautifulSoup4` and `pynliner` in this PR. What are they being used for?

requirements_mysql.txt
```diff
@@ -1,7 +1,7 @@
-mysqlclient==2.0.2 \
-    --hash=sha256:3739824f4b8f29a6a4e5fc965f933fdce7c9edc2454990c3e5c6e038779e5580 \
-    --hash=sha256:4861e33b9fad83dadf38feae77a14106365fc8c354ec4300fab5d8ccfbee4b4e \
-    --hash=sha256:4d91dde55230a512937aafc5981c6c11c0efad2c6b3d33e83287208c8bdab367 \
-    --hash=sha256:8944d9008c29abfa43e3c97551416ea23b2ccb017af85682b79f34fc26dc43ae \
-    --hash=sha256:8df057b08fc27d8f7106bfa997d0a21e2acef017f905f06d6fb0aa6a20d4d2b2 \
+mysqlclient==2.0.3 \
```
davidpofo, line 1: could you reverse this hash and version change. Other than that this PR looks good to me.

HeatherSoron:
@gregelin Anything in `requirements.txt` was handled by the update script (`requirements_txt_updater.sh`), so I'm not sure what those two modules were being used for (looks like dependencies of something else?). However, re-running that script removes those two dependencies, it looks like.

Should I re-run the updater script and then include that in the PR? (I'm not sure what the desired standard procedure is with regards to requirements.txt, and since you and @davidpofo both mentioned something related to that, I figure I'll ask)

requirements_mysql.txt
HeatherSoron: Can do, but it looks like it will get changed again the next time someone runs `requirements_txt_updater.sh`. If you mean "let's not change this now", makes sense to me, so I'll go ahead and revert it; if it's "we want to avoid 2.0.3", that might need a different approach?

peterkaminski: It's "let's not change this now".

We have a convention to run the updater, then trim out any changes that aren't absolutely required by the PR at hand.

(And then periodically / as needed, create a PR which just updates the hashes, but that's separate from regular PRs that change other stuff.)

HeatherSoron: Thanks, that helps clarify procedure.

In that case, I'll check if there's anything needed for this PR that isn't in the file on `develop` branch, and if not, I'll re-merge develop into this PR and use that version of `requirements.txt` verbatim.

HeatherSoron:
Merged in current state of `develop`, this time resolving all `requirements.txt` (etc.) conflicts by using the already-approved version of those files from the main branch.

PR #1258, opened 2021-01-04
Default organizational parameters

woodt:
Support loading of default organizational parameters from JSON files.


woodt:
OK, this is ready for review/merge!

CHANGELOG.md
```diff
@@ -10,8 +10,12 @@ v.999 (December XX, 2020)
 * Support defining multiple allowed hosts via the `local/environment.json` file.
 * Allow administrators to change component name in Component Library.
 * Existing projects can be moved between existing portfolios.
+<<<<<<< HEAD
 * Edit existing portfolio's title and description.
 * Delete existing portfolio.
+=======
```
bama4, line 16: Looks like the merge conflict headers snuck their way into the changelog

controls/models.py
```diff
@@ -620,6 +622,47 @@ def get_baseline_controls(self, baselines_key, baseline_name):
...
+                raise Exception("Duplicate default organizational parameters name {} from {}".format(name, f))
+            self.cache[name] = values
+    
+    def load_param_file(self, path):
+        with path.open("r") as json_file:
+            data = json.load(json_file)
+            if 'name' in data and 'values' in data:
+                return (data["name"], data["values"])
+            else:
+                raise Exception("Invalid organizational parameters file {}".format(path))
```
bama4, line 657: Perhaps there is a more specific exception that can be raised other than the generic `Exception`?

CHANGELOG.md
woodt: Oops.  Will fix this, thanks.

controls/models.py
woodt: What would you suggest, @bama4 ?  Is there a GovReady-Q exception hierarchy I should be using?

gregelin (approved):
Looks great!

controls/models.py
gregelin: Going to carry the generic Exception categorization as technical debut @bama4, @woodt . Do you have an example exception hierarchy we could consider?

PR #1268, opened 2021-01-06
Da/add viewappsource adminaction

davidpofo:


gregelin (approved):
Looks good and tested!

PR #1274, opened 2021-01-08
Copy Components on Import

omarabed15:
For now, while we aren't sure of the behavior for using UUIDs in Component and Statement imports, I have removed the UUID checks from the Component Importer. Instead, we're only checking for conflicts of the Component Name. If there is a conflict, the name gets incremented, so the Component still gets imported and the statements still get created. The only situation where a Statement won't get created now is if the catalog/profile does not contain the control in question.

### DIFF
--- a/controls/utilities.py
+++ b/controls/utilities.py
@@ -91,3 +91,14 @@ def get_control_statement_part(control_stmnt_id):
     # Portion after the '_smt.' is the part
     split_stmnt = control_stmnt_id.split("_smt.")
     return split_stmnt[1] if len(split_stmnt) > 1 else ""
+
+
+def increment_component_name(component_name):
+    """Increments a Component Name by adding (1)"""
+
+    if re.search("\((\d+)\)$", component_name):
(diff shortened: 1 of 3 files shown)


controls/utilities.py
```diff
@@ -91,3 +91,14 @@ def get_control_statement_part(control_stmnt_id):
     # Portion after the '_smt.' is the part
     split_stmnt = control_stmnt_id.split("_smt.")
     return split_stmnt[1] if len(split_stmnt) > 1 else ""
+
+
+def increment_component_name(component_name):
+    """Increments a Component Name by adding (1)"""
+
+    if re.search("\((\d+)\)$", component_name):
+        new_component_name = re.sub("\((\d+)\)$", lambda m: " (" + str(int(m.groups()[0])+1) + ")", component_name)
```
davidpofo, line 100: this reminds me I should learn a lot more regex. Cool stuff

davidpofo (commented):
lgtm. I think test are being made so I will approve after reviewing those too.

omarabed15:
> lgtm. I think tests are being made so I will approve after reviewing those too.

@davidpofo Test updates are now included.

davidpofo (approved):
lgtm ran all tests works as intended

PR #1279, opened 2021-01-11
Da/smt history

davidpofo:


controls/views.py
```diff
@@ -803,6 +803,46 @@ def import_component(request):
...
+    recent_smt = smt.history.first()
+    historical_smt = smt.history.get(history_id=history_id)
+
+    # saving historical statement as a new instance
+    historical_smt.instance.save()
+    logger.info(f"Restoring the current statement with an id of {smt_id} to version with a history id of {history_id}")
+    messages.add_message(request, messages.INFO, f'Successfully restored the statement to version history {history_id}')
+
+    # Diff between most recent and the historical record
+    all_hist = smt.history.all()
```
omarabed15, line 831: Instead of calling `smt.history.all()` twice (about 10 lines later, it is called again) and calling `smt.history.first()` 10 lines prior, a single call to `smt.history.all()` should be made at the start and stored. Then you can just call `.first()` on that variable. So like:
`statement_history = smt.history.all()`
`first_statement_history = statement_history.first()`

Then you can use those variables throughout where needed without re-calling those functions. (I'm not sure how much processing it takes to look up a statement history, but it might end up being a lot down the road as statement histories change a lot. So, removing at least one or two calls to those functions might help!

omarabed15 (changes requested):

controls/views.py
```diff
@@ -793,6 +793,46 @@ def import_component(request):
     result = ComponentImporter().import_component_as_json(oscal_component_json, request)
     return component_library(request)
 
+def statement_history(request, smt_id=None):
+    """Returns the history for the given statement"""
+    from controls.models import Statement
+
+    smt_smt = Statement.objects.get(id=smt_id)
```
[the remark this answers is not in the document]
davidpofo: Good idea i'll except into a message

templates/controls/statement_history.html
```diff
@@ -12,26 +12,24 @@ <h3 class="text-center"> (This statement comes from the {{ statement.0.producer_
...
                 {% for smt_record in statement %}
                     <tr>
-                        <td>{{ smt_record.id }}</td>
-                        <td>{{ smt_record.history_id }}</td>
-                        <td>{{ smt_record.history_date }}</td>
+                        <td class="col-xs-1 col-sm-1 col-md-1">{{ smt_record.history_id }}</td>
+                        <td class="col-xs-2 col-sm-2 col-md-2">{{ smt_record.history_date }}</td>
                         <td>{{ smt_record.body }}</td>
-                        <td><a href="{% url 'restore_to' smt_id=smt_record.id history_id=smt_record.history_id %}" name="restore" class="btn btn-xs btn-primary">Restore</a></td>
+                        <td class="col-xs-2 col-sm-2 col-md-2">{% if forloop.first %} {#This is the current version#} {% else %}<a href="{% url 'restore_to' smt_id=smt_record.id history_id=smt_record.history_id %}" name="restore" class="btn btn-xs btn-primary">Restore to this version</a>{% endif %}</td>
```
omarabed15, line 28: Great catch. I might consider switching around the condition since nothing happens in the "true" clause besides a comment. Could do: `if not forloop.first` and then avoid the `else` all together.

controls/views.py
```diff
@@ -817,26 +822,33 @@ def restore_to_history(request, smt_id, history_id):
     """
     Restore the current model instance to a previous version
     """
+    full_smt_history = None
+    try:
+        smt = Statement.objects.get(id=smt_id)
+        recent_smt = smt.history.first()
+    except:
```
omarabed15, line 829: Just in case multiple exceptions get thrown from lower levels, it's usually helpful to specify the particular exception you're trying to handle (and I think it's something Django Doctor likes to see). When using a "get()", I usually do:
`except Statement.DoesNotExist:`

controls/views.py
```diff
@@ -817,26 +822,33 @@ def restore_to_history(request, smt_id, history_id):
     """
     Restore the current model instance to a previous version
     """
+    full_smt_history = None
+    try:
+        smt = Statement.objects.get(id=smt_id)
```
omarabed15, line 827: Since none of the remaining code really makes sense if this line doesn't succeed (such as subsequent calls using the smt object), I'd put a `return` statement in the `except` clause of this try-except block to terminate execution of this view early, rather than trying to perform the rest of the processing with a null object. Maybe by either copy-pasting this line at the end of the `except`:
`return render(request, "controls/statement_history.html", context)` (and adding placeholders for the context attributes), or reworking the code flow structure somehow. But adding another return statement in the except is probably the lowest effort solution.
Granted, by this point you've already initialized all the variables you are passing out of the view to the template, so any errors in the `try` blocks should just pass the defaults to the template, but I'm talking more about wasting computing and lookup cycles once we already know the code will break. It's a theoretical performance improvement approach, but we can talk about it more outside.

davidpofo: After the first error in a try it would fail when providing invalid history ids and/or when checking the smt object.

controls/views.py
davidpofo: Statement.DoesNotExist keep getting to the yellow error page so I optted for ObjectDoesNotExist from django.core.exceptions

controls/views.py
omarabed15: True, it'll fail either way, so this is fine. I was just thinking to terminate execution of the view early, but it probably isn't much of a performance savings.

omarabed15 (approved):
Looks good!

PR #1285, opened 2021-01-12
Add link to Compliance App Library without adding app to portfolio.

omarabed15:
I've had trouble accessing the list of Compliance Apps without specifying a portfolio, but there should be a way to view available Compliance Apps without adding them to a project or portfolio. I created a simple workflow that removes the "Add" button from the Compliance Apps page if a portfolio is not specified. This should improve the workflow without requiring users to enter a specific URL.


davidpofo (approved):
lgtm

PR #1286, opened 2021-01-12
Plumbing to support OSCAL JSON SSPs

woodt:
* Adds a '%dict" operator to JSON/YAML template rendering.  Like the '%for' operator, but generates dict instead of list.
* Fix caching issue with Catalog.GetInstance() ... when hash was negative, instances were not cached!
* Replace use of Catalog() constructor with Catalog.GetInstance()
* Add an 'oscal' context to JSON/YAML template rendering.
* Add a placeholder route and view for returning an OSCAL profile for a system.


gregelin:
@woodt Great that we have our own copy of the schema. Is the code set up to handle multiple versions of the OSCAL schema? It's entirely possible in the future that some systems require being stuck with a particular OSCAL version.

gregelin (approved):
Looks great!

PR #1287, opened 2021-01-12
Da/smt history

davidpofo:
Added past and new change reason with a text area for the latter.


davidpofo:
Need to implement a url_for_full_content of statement body in the history something close to this
`<td>{{ smt_record.body|safe|truncatewords:"20"|linebreaks }}       <a href="{{ url_for_full_content }}">read more</a></td>
`

PR #1313, opened 2021-01-18
Update install scripts. Add lightweight-ato to default apps.

gregelin:
Add Lightweight-ATO compliance app to the set of default
compliance apps that ship with GovReady-Q to make getting
started even easier.

Update the default `local/environment.json` file recommended by
`first_run` command and written by `install-govready-q.sh` to
use `govready-url` parameter rather than deprecated `host` and `https`
parameters.

By default, set organization name to "main".
Skip creating organization if already exists.

Add optional `PIPUSER` parameter to `install-govready-q.sh`
to avoid error of running pip install with `--user` flag in
virtual environments. In fact, remove the entire testing for
a virtual environment because people can have many different
paths to virtual environment.

Comment out starting GovReady-Q server automatically because
too many edge cases exist to execute that well. Also, not
all bash shells have library jq installed.

Update CHANGELOG.md.


gregelin:
This PR was incorrectly made against `master`. Closing.

PR #1314, opened 2021-01-18
Update install scripts. Add lightweight-ato to default apps.

gregelin:
Add Lightweight-ATO compliance app to the set of default
compliance apps that ship with GovReady-Q to make getting
started even easier.

Update the default `local/environment.json` file recommended by
`first_run` command and written by `install-govready-q.sh` to
use `govready-url` parameter rather than deprecated `host` and `https`
parameters.

By default, set organization name to "main".
Skip creating organization if already exists.

Add optional `PIPUSER` parameter to `install-govready-q.sh`
to avoid error of running pip install with `--user` flag in
virtual environments. In fact, remove the entire testing for
a virtual environment because people can have many different
paths to virtual environment.

Comment out starting GovReady-Q server automatically because
too many edge cases exist to execute that well. Also, not
all bash shells have library jq installed.

Update CHANGELOG.md.

The large number of files changed is from adding the Lightweight ATO app directory.


install-govready-q.sh
```diff
@@ -131,33 +136,44 @@ fi
...
+# 	while true;
+# 	do
+# 		read -p "Do you want to run the GovReady-Q now, in the foreground? [y/n] " answer
+# 		case $answer in
+# 			[Yy]* ) python3 manage.py runserver "$govready_host_and_port"; break;;
+# 			[Nn]* ) break;;
+# 			* ) echo "Please answer 'yes' or 'no'";;
+# 		esac
+# 	done
+# fi
```
omarabed15, line 179: Should probably delete the commented code completely if it's not being used, but other than that, this PR looks great to me.

gregelin: Ok. Makes sense. I'm just afraid I'll forget how something was done...

gregelin:
Capturing here some code using `jq` to parse json in command line to get environmental variables in order to provide an option to start up GovReady-Q.

The following unused code was removed from `install-govready-q.sh` but I want to remember how the code works.

```
# TODO - Need a test to see if jq is installed on system; commented out code in meantime
#
# Use jq and cut to extract hostname from govready-url param in local/environment.json
environment_json=$(<local/environment.json)
govready_url=`cat local/environment.json | jq .\"govready-url\"`
# echo "govready_url is $govready_url"
govready_host_and_port=$(echo $govready_url | cut -f2 -d\" | cut -f3 -d/)
echo "govready_host_and_port is $govready_host_and_port"


# TODO - Need a better way to start GovReady because of different use cases
#        Commenting out code in meantime
#
if [ $NONINTERACTIVE ];
then
  echo "Starting GovReady-Q Server..."
  python3 manage.py runserver "$govready_host_and_port"
else
  # prompt the user if they want to run the webserver now
  # this currently runs synchronously, in the foreground, so it needs to be the last
  # functional line of the script
  while true;
  do
      read -p "Do you want to run the GovReady-Q now, in the foreground? [y/n] " answer
      case $answer in
          [Yy]* ) python3 manage.py runserver "$govready_host_and_port"; break;;
          [Nn]* ) break;;
          * ) echo "Please answer 'yes' or 'no'";;
      esac
  done
fi
```

omarabed15 (approved):

PR #1315, opened 2021-01-18
Test wait for generic

davidpofo:
A bunch of improvements on tests run locally and in circleci.
Includes:

- A configuration file for coverage
- A wait_for_sleep_after function to prevent excessibe var_sleep(s)
- splitting tests via the circleci config.yml to run jobs in parallel


davidpofo:
@omarabed15 I particularly want a mac user to run the full test suite with all these changes to make sure it is OS independent

siteapp/tests.py
```diff
@@ -29,11 +33,22 @@
...
+def wait_for_sleep_after(fn):
+    MAX_WAIT = 10
+    start_time = time.time()
+    while True:
+        try:
+            return fn()
+        except (AssertionError, WebDriverException) as e:
+            if time.time() - start_time > MAX_WAIT:
+                raise e
+            time.sleep(0.5)
```
omarabed15, line 50: Awesome, I like this a lot.

controls/tests.py
```diff
@@ -208,13 +204,13 @@ def test_component_download_oscal_json(self):
         elif os.path.isfile(self.json_download.name):
             os.remove(self.json_download.name)
         self.click_element("a#oscal_download_json_link")
-        var_sleep(2)            # need to wait for download, alas
+        var_sleep(1)
```
omarabed15, line 207: Should all calls to `var_sleep` just be replaced with `wait_for_sleep_after`? If not, I'm just curious why not.

controls/tests.py
```diff
@@ -280,9 +276,8 @@ def test_component_import_oscal_json(self):
 
         self.click_element('input#import_component_submit')
 
-        var_sleep(3) # Wait for OSCAL to be imported
+        element_count_after_import = wait_for_sleep_after(lambda: Element.objects.filter(element_type="system_element").count())
```
omarabed15, line 279: Here, we want to wait for sleep after the previous command (`click_element`), right? Here, I think the wait is being called on the wrong function.

controls/tests.py
```diff
@@ -293,19 +288,18 @@ def test_component_import_oscal_json(self):
         # but we can't test individual statements because the UUIDs are randomly generated and not consistent
         # with the OSCAL JSON file. So we simply do a count.
 
-        var_sleep(3) # Needed to allow page to refresh and messages to render
-
         # Test that duplicate Components are re-imported with a different name and that Statements get reimported
-        self.click_element('a#component-import-oscal')
+
+        wait_for_sleep_after(lambda: self.click_element('a#component-import-oscal'))
```
omarabed15, line 293: This seems to be a recurring question. Here, the wait should happen before this click. Instead, it seems like the element is being clicked and _then_ we're waiting for the wait. If this is the correct use of the function, maybe it just needs to be renamed to "wait_for_sleep_before", or maybe it is a logic issue and the call to the wait function needs to be on the function before the initial var_sleep.

controls/tests.py
```diff
@@ -821,16 +813,16 @@ def test_smt_autocomplete(self):
         search_comps_txtbar[-1].click()
         search_comps_txtbar[-1].clear()
         search_comps_txtbar[-1].send_keys("Component")
-        self.browser.find_elements_by_id("selected_producer_element_form_id")[-1].click()
-        var_sleep(3)
+        self.browser.find_elements_by_id("selected_producer_element_form_id")[-1].click()# Ajax request
+        var_sleep(2)
```
omarabed15, line 817: multiple var_sleep calls here can probably be swapped out, but I'm curious if that's not the case.

siteapp/tests.py
```diff
@@ -266,72 +285,75 @@ def setUp(self):
...
         self.user.user_permissions.add(Permission.objects.get(codename='view_appsource'))
-        self.user.save()
+        wait_for_sleep_after(lambda: self.user.save())
         self.user.reset_api_keys()
         self.user.user_permissions.add(Permission.objects.get(codename='view_appsource'))
         self.client.login(username=self.user.username, password=self.user.clear_password)
 
         # Create a Portfolio and Grant Access
-        portfolio = Portfolio.objects.create(title=self.user.username)
+        portfolio = Portfolio.objects.get_or_create(title=self.user.username)[0]
```
omarabed15, line 317: I'm curious, what's the need for the `[0]` at the end? Shouldn't `get_or_create` only return a single object? Same question throughout this file.

siteapp/tests.py
```diff
@@ -398,17 +416,14 @@ def _accept_invitation(self, username):
         # Extract the URL in the email and visit it.
         invitation_body = self.pop_email().body
         invitation_url_pattern = re.escape(self.url("/invitation/")) + r"\S+"
-        print("invitation_url_pattern", invitation_url_pattern)
+        #print("invitation_url_pattern", invitation_url_pattern)
```
omarabed15, line 419: Can remove commented out print statements throughout.

omarabed15 (changes requested):
Overall, looks good. I have some conceptual questions, and am proposing either a name change or logic change to the function `wait_for_sleep_after`. Maybe `wait_then_run` or `run_then_wait` depending on the intended use of the function.

davidpofo:
> Overall, looks good. I have some conceptual questions, and am proposing either a name change or logic change to the function `wait_for_sleep_after`. Maybe `wait_then_run` or `run_then_wait` depending on the intended use of the function.

yeah I think the name should be consistent since we are waiting for the function and then sleeping if it fails in the try. Then we wait again (while loop) and stops finally when  the time is higher than max_wait

controls/tests.py
davidpofo: I am not sure why but especially file uploads/downloads are fickle and still need the var_sleeps.

controls/tests.py
davidpofo: So the wait_for is applied to allow whatever happened directly after a var_sleep to occur without issue. So if it needs more than 3 or less than 3 this allows in this case element_count_after_import  to still be created.

siteapp/tests.py
davidpofo: get_or_create  returns a tuple with the 0 index being the actual portfolio in this case

controls/tests.py
davidpofo: So if self.click_element can't find the element to click this should fail the first time. If it does the wait_for_sleep_after func will recall this on that element for up to 10seconds(max_wait). So it is not clicking several times, it is instead _attempting_ to click the element however many times.

controls/tests.py
davidpofo: yeah ajax like file uploads/downloads are edge cases that force the need for var_sleeps

controls/tests.py
omarabed15: I see. I think the phrasing of the function threw me off. It initially had me thinking "Wait for sleep after doing the param function" which is backwards. In that case, I think "wait before" would be more accurate, correct? What about something like "wait_then_execute"? With the combo of the function name as the parameter, it might read intuitively, like: `wait_then_execute(some_function)`

controls/tests.py
omarabed15: Okay great, that makes sense. So it's really a naming misunderstanding on my part. I was thinking the function said "wait for sleep _after_ doing this thing". But it's really saying "wait for sleep, then do this thing afterwards". See my previous naming suggestions to maybe clarify this function, but the actual logic of it seems good to me.

omarabed15 (approved):
Perfect, in that case everything looks good logic-wise. I would recommend a simple naming change to the wait function for clarity, but that's all I can say.

PR #1323, opened 2021-01-20
Update OSCAL component import/export, tests, and schemas for 1.0.0-rc1

woodt:


controls/views.py
```diff
@@ -529,7 +529,7 @@ def create_component(self, component_json, request):
         @returns: Element object if created, None otherwise
         """
 
-        component_name = component_json['name']
+        component_name = component_json['title']
```
omarabed15, line 532: Nice, this was the only concern from my end. Good catch.

omarabed15 (approved):
Looks good to me! I also checked the tests, and they don't seem to rely on the old schema, so I think this is good!

davidpofo (approved):
lgtm

PR #1334, opened 2021-01-21
Oscal app inputs simple

omarabed15:
**Compare against:** https://github.com/GovReady/govready-q/pull/1335
**ONLY MERGE ONE**

This is a simple approach to supporting OSCAL JSON components with Compliance App sources.
Iterates through inputs in the app.yaml file.

Input format in the app.yaml file is expected as:
```
input:
- id: oscal_components
  name: OSCAL Components
  type: oscal
  path: data/oscal_components.json
  group: OSCAL
```

PR #1335, opened 2021-01-21
Oscal app inputs prop

omarabed15:
**Compare against:** https://github.com/GovReady/govready-q/pull/1334
**ONLY MERGE ONE**

This approach to supporting OSCAL JSON components with Compliance App sources saves the input files in the database, though I'm not sure if that is desired/necessary.
Iterates through inputs in the app.yaml file.
Input format in the app.yaml file is expected as:
```
input:
- id: oscal_components
  name: OSCAL Components
  type: oscal
  path: data/oscal_components.json
  group: OSCAL
```


omarabed15:
Superceded by #1336

PR #1339, opened 2021-01-22
 Allow migrations to be ran backwards

django-doctor:
currently django would raise exception if we tried to reverse the migrations because the data migration has no backwards. This avoids that problem [read more](https://django.doctor/advice/C8001)

I made this PR via the "auto fix" tab on https://django.doctor/GovReady/govready-q

### DIFF
--- a/controls/migrations/0023_auto_20200802_0119.py
+++ b/controls/migrations/0023_auto_20200802_0119.py
@@ -21,19 +21,24 @@ class Migration(migrations.Migration):
             obj.save()
 
     operations = [
-        migrations.RemoveField(
-            model_name='poam',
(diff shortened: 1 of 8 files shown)


gregelin:
@davidpofo Any thoughts on how we make this PR against `develop` instead of `master`? Ore do you want to merge into `master`?

davidpofo:
> @davidpofo Any thoughts on how we make this PR against `develop` instead of `master`? Ore do you want to merge into `master`?

I think that's a setting you can make as the repository owner if you go to their website. Also can just change the base on this page if you click 'edit' in the top right --> select develop

davidpofo (approved):

PR #1356, opened 2021-01-31
Display lightweight-ato SSP button under action-buttons

gregelin:
Update included lightweight-ato so a button to navigate
to the SSP always displays under the action button.

First, adjust lightweight-ato/app.yaml ssp_intro queestion
to always display rather than the display be imputed.
Next use `placement: action-buttons` to display button
underneath action-buttons instead of within the group blocks.
Finally, create a css class for a putting the title of
of the compliance app above the action buttons.
Update style of button.

Remove the "Upgrade App" button since that feature is now
better done within the settings page.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,10 +9,20 @@ v999 (January XX, 2021)
 **UI changes**
 
 * Remove "Upgrade Project" button from project page action buttons. Upgrade is now in settings page.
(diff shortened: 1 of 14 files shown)


kerryrm (approved):
Seems fine.

davidpofo (approved):
lgtm

PR #1357, opened 2021-01-31
Ge/UI store

gregelin:
Improve styling of app store items.

Give each app store listing a clean, gray box
Clean up "Add" and "Info" buttons.

Display "Unknown" when app vendor is set to "None"
instead of "none".

Clean up format style sheets in project.html, app-store.html,
app-store-item.html templates. Properly indent local
project.html template style sheets to collapse in IDEs.

Tweek styling of project page question page:

* Remove light gray background from project page,
  question page, task finished page.
* Reduce corner radius in focus area blocks.
* Widen question area.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,14 @@ v999 (January XX, 2021)
 
 [Add updates here]
 
+**UI changes**
+
+* Remove "Upgrade Project" button from project page action buttons. Upgrade is now in settings page.
+
+**Compliance app changes**
+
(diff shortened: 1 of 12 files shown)


kerryrm (approved):
I would set border:1 instead of 3, but looks ok (.app). It also seems like the icon/graphic should be on the left, but that can be changed at a later date. A revised store probably needs to be mocked up and discussed. These changes move things forward for now. The use of px instead of em may prove problematic but again, can be adjusted later.

davidpofo (approved):
lgtm

PR #1362, opened 2021-02-01
Da/element details content bug

davidpofo:
should address import of project issue. also some element details items. also added functools.lru_cache() decorator to speed a couple funcs.


gregelin (approved):
Looks good

PR #1371, opened 2021-02-04
Faster and better caching, pagination and more

davidpofo:
Replaced the function-based view components_selected with class-based listview SelectedComponentsList and ProjectList. The latter also has pagination and is ordered by date created. Some caching and optimizations.


gregelin:
@davidpofo Can you check my updates to `CHANGELOG.md` for completeness, please?

siteapp/settings.py
```diff
@@ -136,6 +136,10 @@ def make_secret_key():
 	'simple_history.middleware.HistoryRequestMiddleware',
 	'pyinstrument.middleware.ProfilerMiddleware',
 ]
+CACHE_MIDDLEWARE_ALIAS='default'
+CACHE_MIDDLEWARE_SECONDS=604800
```
gregelin, line 140: @davidpofo Can you add some comments regarding what this cache does?

siteapp/settings_application.py
```diff
@@ -1,10 +1,9 @@
-# nothing here
-
 import re
+import sys
 from .settings import *
 
 INSTALLED_APPS += [
-    #'debug_toolbar',
+    'debug_toolbar',
```
gregelin, line 6: @davidpofo Do we want Debig toolbar running all the time?

gregelin (changes requested):

siteapp/settings_application.py
davidpofo: Added a new enable_toolbar field in environment.json/settings to turn this on/off.

siteapp/settings.py
davidpofo: I just removed as this was an attempt at site-wide caching that is more trouble than it is worth

gregelin: I have an idea on improving performance. Let's discuss after the rATO standup?

siteapp/settings_application.py
gregelin: Please remember to `enable_toolbar` param to documentation: https://govready-q.readthedocs.io/en/latest/installation-and-setup/configuration-with-environment-variables.html

davidpofo:
PR party 2/11. Future enhancements. @kerryrm not sands-serif it is san-serif.

PR #1372, opened 2021-02-04
Fix `requirements.txt` < `requirements_txt_updater.sh`  to prevent install error  re: setuptools not pinned

Dav1dV:
when  `install-govready-q.sh`  /  `pip3 install -r requirements.txt`

  > In --require-hashes mode, all requirements must have their versions pinned with ==. These do not:
  >     setuptools from

---

Fix `requirements_txt_checker.sh`  to prevent false "not in sync" failure

  when  `requirements.txt` contains  either of the following:
  - multiline annotations  via  `pip-compile --generate-hashes` or `--annotate`
  - comments               via  `pip-compile --allow-unsafe`

  > 'requirements.txt' is not in sync with 'requirements.in'. Some packages may have updates available. Run requirements_txt_updater.sh.

### DIFF
--- a/requirements_txt_checker.sh
+++ b/requirements_txt_checker.sh
@@ -20,7 +20,9 @@ function run_checks() {
 	# a temporary file.
 	FN=$(mktemp)
 	echo "Flattening transitive dependencies of '$FILE_BASE.txt' to a temporary file '$FN'"
-	pip-compile --generate-hashes --output-file $FN --no-header --no-annotate ${FILE_BASE}.in > /dev/null
+	pip-compile --generate-hashes --allow-unsafe  --upgrade --output-file $FN --no-header --no-annotate ${FILE_BASE}.in > /dev/null
+        # like requirements_txt_updater.sh
+        #   except added --no-annotate (see below)
 
 	# The reverse-dependency metadata doesn't seem to be entirely
 	# accurate and changes nondeterministically? We omit it above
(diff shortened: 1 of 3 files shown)


gregelin:
@davidpofo Do you have thoughts on this? @peterkaminski

davidpofo:
> @davidpofo Do you have thoughts on this? @peterkaminski

Looks good to me tested it on my wsl ubuntu.

requirements_txt_checker.sh
```diff
@@ -29,8 +31,37 @@ function run_checks() {
...
+	#   >     #   importlib-resources
+	#   >
+	#   > # The following packages are considered to be unsafe in a requirements file:
+	#   > setuptools==53.0.0 \
+	#   >     --hash=sha256:0e86620d658c5ca87a71a283bd308fcaeb4c33e17792ef6f081aec17c171347f \
+	#   >     --hash=sha256:1b18ef17d74ba97ac9c0e4b4265f123f07a8ae85d9cd093949fa056d3eeeead5
+	#   >     # via
+	#   >     #   fs
+	#
+	#   to
```
gregelin, line 53: @Dav1dV I am a bit confused here. Are we losing the"via" comments? I find the "via" comments very useful on transitive dependencies.

Dav1dV: No - this is fixing `requirements_txt_checker.sh`'s preexisting temp `requirements.txt` de-annotation
_(e.g., to support newer multiline `# via\n    #   ...` `pip-compile` annotations)_
for comparison w/ the also-temp `pip-compile --no-annotate` output generated above
per the preexisting comment above:
> ... remove it from a copy of our requirements.txt
> file before comparing

P.S.  Agree about keeping the `# via ...` `pip-compile` annotations.

davidpofo:
@Dav1dV Will address concerns PR party 2/11

PR #1411, opened 2021-02-12
Additional logging

gregelin:
Adds in some additional logging.

NOTE: Looking to merge older branch. Do a careful review.


siteapp/settings.py
```diff
@@ -453,6 +453,9 @@ def make_secret_key():
 	TEMPLATES[0].setdefault('DIRS', [])\
 		.insert(0, os.path.join(environment["branding"], 'templates'))
 
+# Load logging configuration from settings_logging.py.
+from .settings_logging import *
```
bama4, line 457: We may want to import the specific packages that we need instead of using `*`. This will make the code and dependencies more readable.

gregelin:
Overtaken by events.

PR #1414, opened 2021-02-13
Statement history permission checks fix & tweaks

### LINKED ISSUE #1300: Restrict viewing statement history to appropriate users
**Is your feature request related to a problem? Please describe.**
Anonymous users should not be able to view any details about a system or a system's control statement history; only logged in users with authorization to view a system should be able to view system's control statement history.

I think the permissions for statement history for a component in the library are slightly different. The default might be that all users can see the library of components (including anonymous users) but that some components are "private" visible only to individual users and groups of users. 

**Describe the solution you'd like**
- Consider adding `@login_required` decorator to certain views
- Check user has permission to view system
- Permissions to view statement history
- Only a user with edit privileges on a component library statement should be able to revert to a previous version (e.g., author of the component library, or admin)

Dav1dV:
- Fix statement history getting & version restoring
  - Resolve #1300

### DIFF
--- a/controls/views.py
+++ b/controls/views.py
@@ -836,60 +836,49 @@ def import_component(request):
     result = ComponentImporter().import_components_as_json(import_name, oscal_component_json, request)
     return component_library(request)
 
+
+def raise_404_if_not_permitted_to_statement(request, statement, system_permission='view_system'):
+    """Raises a 404 if the user doesn't have the statement system permission"""
+    while True:
+        if request.user.is_superuser or request.user.is_staff:
+            # Allow access if the user has superuser or staff member permissions respectively
+            break
+        else:
+            # Allow access if the user has the statement system permission
+            try:
+                system = System.objects.get(root_element=statement.consumer_element)
+                if request.user.has_perm(system_permission, system):
+                    break
+            except System.DoesNotExist:
+                pass
+
+        raise Http404
+
+def raise_404_if_not_permitted_to_change_statement(request, statement):
+    """Raises a 404 if the user doesn't have permission to change the statement"""
+    raise_404_if_not_permitted_to_statement(request, statement, 'change_system')
+
 @login_required
 def statement_history(request, smt_id=None):
-    """Returns the history for the given statement"""
+    """Returns the history for the specified statement"""
 
-    from controls.models import Statement
     # Get statement if exists else 404
-    smt = Statement.objects.get_object_or_404(id=smt_id)
+    smt = get_object_or_404(Statement, id=smt_id)
 
-    # Check permission block
-    permission = False
-    if request.user.is_superuser or request.user.is_staff:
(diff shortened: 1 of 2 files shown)


davidpofo:
Restoring version fixed is already addressed in #1413 It was simply a miss use of get_object_or_404 on my part went from smt = Statement.objects.get_object_or_404(id=smt_id) to get_object_or_404(Statement, id=smt_id). @Dav1dV Could you remove that part of this PR as it is a dupe.

Dav1dV:
> Restoring version fixed is already addressed in #1413 It was simply a miss use of get_object_or_404 on my part went from smt = Statement.objects.get_object_or_404(id=smt_id) to get_object_or_404(Statement, id=smt_id). @Dav1dV Could you remove that part of this PR as it is a dupe.

@davidpofo Git will skip same changes in multiple PRs for us;
however,  if I revert that fix in this PR,  it will break this PR & `develop` if this PR is merged after #1413

I did remove the unnecessary import line 843 above like you did in #1413
so it won't matter what order the PRs are merged in.
https://github.com/GovReady/govready-q/pull/1414/files#diff-f51ba1da4e54cfb61d01e6b0612e9a6ee3b213207448a0953e0c93342ea29b13L843

_Note:  That's for the history getting `statement_history` fix;
the version restoring `restore_to_history` fix is for line 885
which tries to use `smt` before it is assigned:
https://github.com/GovReady/govready-q/pull/1414/files#diff-f51ba1da4e54cfb61d01e6b0612e9a6ee3b213207448a0953e0c93342ea29b13L885_

controls/views.py
```diff
@@ -836,80 +836,65 @@ def import_component(request):
...
+                if request.user.has_perm(system_permission, system):
+                    break
+            except System.DoesNotExist:
+                pass
+
+        raise Http404
+
+def raise_404_if_not_permitted_to_change_statement(request, statement):
+    """Raises a 404 if the user doesn't have permission to change the statement"""
+    raise_404_if_not_permitted_to_statement(request, statement, 'change_system')
```
davidpofo, line 859: Since this is a one-line function I think it would make this functionality cleaner to just use `raise_404_if_not_permitted_to_statement` directly instead of nesting it in this function.

controls/views.py
```diff
@@ -836,80 +836,65 @@ def import_component(request):
...
 
     try:
+        # Save historical statement as a new instance
         historical_smt = smt.history.get(history_id=history_id)
-        # saving historical statement as a new instance
         historical_smt.instance.save()
+
         # Update the reason for the new statement record
-        update_change_reason(smt.history.first().instance, change_reason)
+        recent_smt     = smt.history.first()
```
davidpofo, line 896: nice

davidpofo (approved):

PR #1423, opened 2021-02-18
Set control catalog, baseline on compliance app start based on params in compliance app

gregelin:
GovReady will now read compliance app catalog.parameters.catalog_key and catalog.parameters.baseline values to set the selected controls for a system.

The large number of files change reflects creating a copy of LATO included compliance app with a different default catalog and baseline.


davidpofo (approved):
lgtm

templates/systems/sar_list.html
```diff
@@ -52,6 +52,11 @@
         </div>
       {% endfor %}
     {% endif %}
+<<<<<<< HEAD
+
+  </div>
+=======
+>>>>>>> develop
```
peterkaminski, line 59: unaddressed merge

peterkaminski (dismissed):
needs a merge addressed in `templates/systems/sar_list.html`

is there a way to check all the startpack files? do we need to?

otherwise lgtm

PR #1434, opened 2021-02-25
Added and updated favicons

kerryrm:
Added images for non-browser favicons and updated the browser favicon. No need to update the release notes. No UI changes.


templates/base.html
```diff
@@ -13,8 +13,8 @@
         {% endcomment %}
         <link rel="shortcut icon" href='{% static "img/brand/favicon.ico" %}' type="image/x-icon">
         <link rel="icon" href='{% static "img/brand/favicon.ico" %}' type="image/x-icon">
-        {% comment %}replace with image and HTML generated by http://realfavicongenerator.net/ one day{% endcomment %}
```
davidpofo, line 16: @kerryrm What was this originally supposed to do?

kerryrm: That wasn't my comment. It had been there for years.

PR #1439, opened 2021-02-27
fixing sorting and adding natsort to requirements(in/txt)

kerryrm:
using natsort package to sort SID correctly, taking into consideration letters and numbers. 
Needed to add natsort package to requirements. https://pypi.org/project/natsort/
requirements.txt and .in updated.


kerryrm:
to be clear, this only fixes sort within component_library_component on the components/element_detail_tabs.html page, which was the worst culprit. There are (likely) other places sort will need to be fixed.

controls/views.py
```diff
@@ -43,7 +43,7 @@
 structlog.configure(logger_factory=LoggerFactory())
 structlog.configure(processors=[structlog.processors.JSONRenderer()])
 logger = get_logger()
-
+from natsort import natsorted, ns
```
davidpofo, line 46:
```suggestion
from natsort import natsorted
```

gregelin (approved):
Looks great!

PR #1441, opened 2021-02-27
Fixed component library name/description edit

davidpofo:
Editing components in Component Library fails because if you try to update the description of a component, the check Joao wrote to avoid name collisions is causing the description to not change b/c of an existence of current name. Name collision is wrongly seeing current name as a collision

Created a ElementEditForm Django form in conjunction with some functional changes to avoid name collisions issues with component library.


gregelin (commented):
Looks very good!

gregelin:
@davidpofo There's a conflict that I felt it would be better for you to resolve.

Issue #674, opened 2019-07-26
checksum changed for siteapp/static/vendor/google-fonts.css


peterkaminski:
`fetch-vendor-resources.sh` does checksums against the assets it downloads.

At commit `2dc0e22271af33d66e3f52428f8b07733b414d7a`, the checksum for `siteapp/static/vendor/google-fonts.css` was found to have changed from:

`d36dbb50dcb3932ce7e3a0271d29eb6d95b82efd6ba08044f2570d9e337f5a88`

to:

`990e7373d100faee6fa7d92c1277695520e3c502f726bb28ce03d2b6d2cd3e6c`

This is probably because of a benign update, but it could also indicate introduction of a vulnerability.  For instance, custom web fonts have been known to be used for infiltration, e.g., https://www.bleepingcomputer.com/news/security/new-phishing-tactic-uses-custom-web-fonts-to-prevent-detection/

To clear this issue:

* check that the incoming CSS file, and its linked WOFF/WOFF2 files, are clean
* check that the `google-font-download` script that is used to download the assets is clean

Also, consider archiving old versions of the assets downloaded by `fetch-vendor-resources.sh`, to enable easy `diff` runs against old and new versions of the assets.

gregelin:
Issue has been addressed.

PR #1444, opened 2021-02-27
Better google font download, fix #1443

### LINKED ISSUE #1443: `fetch-vendor-resources.sh` fails on macOS
**Describe the bug as a Story**
`fetch-vendor-resources.sh` fails on stock macOS. This is because `fetch-vendor-resources.sh` downloads and runs [neverpanic/google-font-download](https://github.com/neverpanic/google-font-download) which depends on, and checks for, Bash 4+.  On stock macOS, bash will never be higher than 3.2, due to Apple's decision about GPL licensing.

(A workaround is for the user to install Bash 4+ with, for example, Homebrew. But this complexifies system requirements and setup for the user.)

**To Reproduce**
1. On macOS that doesn't have a non-stock bash, run `fetch-vendor-resources.sh` (or one of the other scripts that runs `fetch-vendor-resources.sh`).
2. After a number of lines of output, observe this line and the related progess hash characters:
```
https://raw.githubusercontent.com/neverpanic/google-font-download/d7bb34e0c5fa01d9411213fd1a1363317d3ceb11/google-font-download...
```
3. Observe this error: "Error: This script needs Bash 4.x to run." (It is output by the google-font-download script.)

**Expected behavior**
* On stock macOS, `fetch-vendor-resources.sh` runs to completion without error.
* After `fetch-vendor-resources.sh` runs, there are Google Font files in `siteapp/static/vendor` (Hind_* and Lato_*)

**Desktop (please complete the following information):**
 - OS: macOS 11.2.1 Big Sur (Intel)
 - Browser: Chrome "Version 88.0.4324.192 (Official Build) (x86_64)"
 - Version: `develop` branch, 678b98c3039d8d4a5f8ca7373e1cb0abf239928f
 - Virtual Machine: no
 - Shell: zsh

## Developer hints

Will fix with `pk-better-google-font-download-20210227` branch.

## Acceptance criteria 

GIVEN stock macOS,
WHEN I run `fetch-vendor-resources.sh`
THEN it runs to completion without error,
AND THEN there are Google Font files in `siteapp/static/vendor` (Hind_* and Lato_*).

peterkaminski:
See #1443 for the problem to fix.

Fix by:

Rather than download the `google-font-download` helper script and run it, just download the font files directly.

Put the `google-font.css` file the helper script usually generates alongside our CSS file in `siteapp/static/css`, since it is no longer generated by the helper script.


gregelin:
@peterkaminski Thanks for the fix to avoid bash4 on Mac. I'm also wondering about simply not using the Google fonts?

gregelin:
@kerryrm How much value do you think the Google fonts Hind_400 and Lato are bringing to GovReady? Peter has invested in figuring out how to simplify the download of these fonts, but I'm wondering if the long term maintenance of these assets is worth the value we are getting from the fonts. Thoughts?

peterkaminski:
@gregelin, the cost of the fix is already sunk, and I don't anticipate that Google will make any changes that will cause us long-term maintenance costs in the foreseeable future.  I believe the static font downloads will stay the same for the foreseeable future.

As background, Lato and Hind are used as the primary header and body fonts throughout the whole product.

Also to note, with this PR, the downloads are now of static assets, rather than executable code, and are just like all the other assets that `fetch-vendor-resources` downloads and checks.

So, with marginal additional cost, basically the acceptance of this PR, I see these benefits in accepting this PR:

* "look" of the current product stays the same as it has been
* a subtly distinctive "look" of chosen fonts for a web product, rather than using the same fonts everyone else uses, is a visual design / branding advantage
* using the pattern in this PR to install these particular Google Fonts documents (by example) the method to set the primary header and body fonts to any other Google Font, for both internal and external developers, in the future

gregelin:
@peterkaminski Thanks for your thoughts. I was less worried about sunk comments than potential ongoing maintenance:

- We previously had to change the google fonts because of a checksum change
- It's one more asset from a remote source that needs to be managed for a deployment into a network that is not connected to the public internet.

Your points regarding a distinctive look and reusing the pattern both resonate.

Would it make more sense to _permamently_ include the fonts in the `siteapp/static` assets directory instead of retrieving them each time? Yes, it makes the deploy larger, but permanently including avoids the need to transfer assets with each deployment.

Come to think of it, maybe we should be managing static assets in a second repository and periodically including all the static assets inside GovReady so GovReady does not need to fetch static assets. Then the static asset library would already be organized as a package of content to move around with the GovReady.

Thoughts?

gregelin:
I'm going to merge this PR, but let's continue to discuss a bit further before closing the issue, @peterkaminski.

peterkaminski:
@gregelin,

> We previously had to change the google fonts because of a checksum change

I think that was a change in the script and its output; but I'm not sure that the fonts themselves changed.  (I didn't look in depth, but for reference, see #674.)

Also note, the fonts are retrieved from URLs that have an apparent version number `/v11/`, `/v17/` in them, which implies those particular files may not change even if they make changes to the fonts at large.

> Come to think of it, maybe we should be managing static assets in a second repository and periodically including all the static assets inside GovReady so GovReady does not need to fetch static assets.

Yes, I think that's a good idea.  The main additional thing that would be required is including copyright statements for the various files.

In the case of these fonts, I did a little looking at this in case we wanted to fix the issue that way instead of continuing to download them. They're both SIL Open Font license.  From a quick read, you would need to take the [boilerplate OFL.txt file](https://scripts.sil.org/cms/scripts/page.php?site_id=nrsi&id=OFL), and customize it with the names, dates, and reserved font name data from [Google Fonts \- Attribution](https://fonts.google.com/attribution).

Something similar would be needed for all the third-party assets.

PR #1445, opened 2021-02-27
Component library and statements enhancement

davidpofo:
Add Sort, Filter, and Search Capability to Component Library


davidpofo:
Note the sorting is dependent on the natsorted package which is added in #1439

gregelin:
@davidpofo These changes look great. Search works. I couldn't figure out a way to get back to the full list of controls after doing a search (e.g., search reset). I tried:
- deleting the values in the search box.
- deleting the values in the search box and hit search again.
- reloading the page

Because the search uses a get (which I like because it is an easy link to share), it is currently necessary to remove the `?search=value` before the page returns all the items.

gregelin:
@davidpofo I think we need an updated `CHANGELOG.md` for this PR?

gregelin:
@davidpofo I committed a change to fix the styling of the pagination span.

davidpofo:
> @davidpofo These changes look great. Search works. I couldn't figure out a way to get back to the full list of controls after doing a search (e.g., search reset). I tried:
> 
> * deleting the values in the search box.
> * deleting the values in the search box and hit search again.
> * reloading the page
> 
> Because the search uses a get (which I like because it is an easy link to share), it is currently necessary to remove the `?search=value` before the page returns all the items.

You could click on the component library button on the nav or enter an empty search in the search box for components. Reloading the page wouldn't reset anything because it would still maintain the page and the search query.

davidpofo:
Added an explicit reset button for the search of components

PR #1456, opened 2021-03-07
Remove deprecated deployment content

gregelin:
Move various deployment directory content out of this repository and into
https://github.com/GovReady/govready-resources repository for long term archive and reference.

This content is no longer being used.


gregelin:
Should be making PR to develop.

PR #1461, opened 2021-03-09
Fix Postgres crash error by setting the ProjectAsset Model content hash length to 128 characters.

gregelin:
Fix Postgres crash error by setting the ProjectAsset Model content hash length to 128 characters.


VERSION
```diff
@@ -1,2 +1 @@
-v0.9.2.1
-
+v0.9.2.2
```
davidpofo: This PR should not be pointing to main. VERSION should be v999.

davidpofo (approved):
lgtm after fixing whats being pointed where and what version this branch should be on.

PR #1462, opened 2021-03-10
component library statement add

davidpofo:
Fixing the adding of a statement functionality


guidedmodules/migrations/0055_auto_20210309_1503.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 3.0.11 on 2021-03-09 15:03
```
gregelin, line 1: What is this migration addressing? What does "inputs linked to this pack" mean?

guidedmodules/models.py
```diff
@@ -120,7 +120,7 @@ class AppVersion(models.Model):
     version_number = models.CharField(blank=True, null=True, max_length=128, help_text="The version number of the compliance app.")
     version_name = models.CharField(blank=True, null=True, max_length=128, help_text="The name of this version/release of the compliance app.")
 
-    input_files = models.ManyToManyField('guidedmodules.AppInput', blank=True, null=True, help_text="The inputs linked to this pack.")
+    input_files = models.ManyToManyField('guidedmodules.AppInput', blank=True, help_text="The inputs linked to this pack.")
```
gregelin, line 123: Are we sure we want to remove `null=True`?

gregelin (changes requested):
This PR needs to be synchronized with https://github.com/GovReady/govready-q/pull/1459 where there are also changes to the 3 column layout including reposition of the search box and buttons, and styling.

guidedmodules/models.py
davidpofo: Yes. It is not used. That is the point of the migration

davidpofo:
> This PR needs to be synchronized with #1459 where there are also changes to the 3 column layout including reposition of the search box and buttons, and styling.

I am not going to synchronize with a PR. Only sync with develop which will include merged & tested PRs. Its develop --> branch off it --> merge to develop. Not branch here, branch sync with other branch then mere to develop.

PR #1464, opened 2021-03-10
updates related to pandoc docx display, SSP generation, SSP templates.

kerryrm:
- new default ssp word docx template
- yaml for SSP cover in docx
- edits to pandoc command
- updates related to pandoc docx display, SSP generation, SSP templates.
- removed markdown, plain text and ODT from options for download
- Merge branch 'develop'  of https://github.com/GovReady/govready-q into km/ssp-layout


gregelin (approved):
Looks great!

gregelin:
@kerryrm Can you add your notes about the change to the `CHANGELOG.md` file, please?

kerryrm:
Notes regarding Pandoc, SSP generation and the template files.

- Pandoc doesn't support colspan in tables. A lot of the template files and yaml files include colspan and this is why tables fail to display properly in a lot of docx exports. The associated files will need to be updated/fixed.

- As of my push to develop, the default docx template is "assets/default-ssp-template.docx" which has the appropriate word styles for an SSP, but the export doesn't have a way to differentiate between the types of exports, so all docx files generated by pandoc will use the SSP template, which is not ideal.
- 	Someone (not me), needs to update the way the code works to have multiple pandoc commands, depending on the type of file being generated/exported. (I can help explain, but the change is beyond my abilities)
	
- Right now, a yaml file is being passed to pandoc which contains SSP specific title page information. This is great for an SSP, but not so great for non-ssp. This needs to be handled differently in the models.py, but again, I don't know how to code this.  Theoretically, there should be different yaml files for different docx files. (--metadata-file=assets/ssp-cover.yaml)

- I am hiding the plain text and markdown export options. They are still "available" if you just click on HTML and change the URL bit from html to markdown. Honestly, those two are of dubious value to a user, particularly since the markdown is a mess. Same goes for ODT. Based on my research, we might want to consider making ODT the default output instead of DOCX, due to the open source nature of ODT and the fact that Word can read ODT just fine. Something to consider down the line.
- 
- Because of the need to make docx export work and look good, the HTML "export" of SSP_v1 does not have a cover page. It's more like a "preview" than an actual export. 
- 	Also, because of the way it is generated/displayed in django, it doesn't technically exist as a file, so a user can't do a "save page as" because they won't actually get the SSP, they will get the front page of GovReady-Q. This is not ideal. That said, I suggest we change the "HTML" export link to just say "Preview" which makes more sense from a user perspective. There's no real value to having an HTML SSP anyway, beyond being able to preview how your questionnaire responses look.
- 	
- I have added code to both the HTML and the DOCX to keep tables from being broke across page breaks. This sometimes works and sometimes doesn't. I don't know what the deal is, but I just wanted to note it.
- 
- -All of my changes are only related to the 800-171 SSP docx generation. There are a lot of other files for other documents that will need to be edited/updated, particularly with regard to colspan and malformed table code (errant <tr>, etc.), but also to accommodate the focus on docx export (with html preview). Consider the changes in this push/branch to be building blocks for future.
- 
- Per greg's request, the metadata yaml file that helps create the title page doesn't include any dynamic information at the moment. Eventually it should be moved out of /assets and made to be part of the questionnaire. That way a user could decide the title and whatever content needs to go on the title page. Of course, the path in models.py will need to be updated. It should also be dynamic since it will be a different file for each type of export (800-171, 800-53, etc).

davidpofo (approved):
LGTM PR party 3/11/21 Kerry mentioned odt will be readded at a later time. default-ssp-template is the name for now (will be dynamic).

PR #1467, opened 2021-03-11
 Polish 3-column statement layout, add col headings

gregelin:
Polish 3-column statement layout, add col headings
    
    Update 3-column statement layout's "edit" into a glyphicon pencil
    pulled all the way right, remove extra lines and other small changes.
    Update 3-column statement layout to include column headings.
    Conditionally display remarks in component library using HTML details tag.


davidpofo (approved):
lgtm PR party 3/11/21

PR #1474, opened 2021-03-13
(WIP) Install docs and changes for Windows compatibility

peterkaminski:


alexanderward:
The backend we use for storing files depends on a binary libmagic ...which only applies to posix based systems like Linux/Mac/Windows-Ubuntu-VM (if they have it).  Suggesting we don't move forward with Windows deployments and focus more on Docker containers if Windows is the deployment environment.

peterkaminski:
Closing, superceded, per @alexanderward, "Suggesting we don't move forward with Windows deployments and focus more on Docker containers if Windows is the deployment environment."

PR #1475, opened 2021-03-13
Add tagging to models starting with components.

gregelin:
This PR adds tagging to components. The tagging is useful to support endusers who are drafting component library content to add labels/tags to the components to group components together. To identify which components came from what teams, which components were auto generated, which components were manually written, etc.

The tagging of components is distinct from admins assigning roles to components. Tags are more temporary and assigned by users. Roles are more permanent, formal, only assigned by admins and connect components to questionnaire actions.

The modification takes a generalized approach to tagging so the tag model can be later added to other models.

Added tag models, views, urls, migrations for reusability.

Add tags to available to models via model mix-in starting with Controls.models.Element.

Display component tags in component library and project components.

Merge component tag related migration.


davidpofo (commented):
all the png files dont seem relevant to this PR. PR party determined this needs to be clearer on value.

gregelin:
@davidpofo I've removed the png images that were unintentionally included in the PR and written more about the benefit of this PR.

davidpofo:
lgtm Future features:adding a filtering by tag. Need to use add/remove tags in UI.

davidpofo (approved):

PR #1489, opened 2021-03-17
changes to and related to the home user page

kerryrm:
updated views.py to allow project data to be surfaced. Changed display of page. Moved CSS to govready-q.css. New icons.


gregelin:
This is what I'm seeing:

Screen Shot 2021-03-17 at 8 23 06 PM

I liked the side-by-side boxes and everything fitting more or less above the fold. 
Also really want to be able to click to go to the projects page, the portfolio page. 

First iteration should not contain listings, me thinks. Current code I think would list ALL projects and all portfolios.

kerryrm:
Yes, I changed the design/layout to put projects higher in the hierarchy. The side by side boxes give everything equal weight, which I think is confusing to a user. I can make projects and portfolios take you to their respective pages, but the idea was to sort of make that unnecessary for a user. If this is where you're sending me after I log in, this is where I want to be able to find the project I've been working on and go there with minimal friction.

Most (normal) users won't have too many projects/portfolios. (could add a "more" to expand the div to show full project list if it exceeds 7 or something)

We can discuss further in other channels.

kerryrm:
(this was also simply meant as a temporary stop-gap, not a permanent design solution)

gregelin:
Overtaken by events.

PR #1495, opened 2021-03-21
project export fails because uuids are not json serializable

davidpofo:
Error occurs when exporting a project with poams.  Poams have a uuid field in their statements and the uuid needs to be typecasted into a string since uuids are not json serializable.


gregelin (approved):
Looks good

gregelin:
@davidpofo Did you test the import with this export?

davidpofo:
> @davidpofo Did you test the import with this export?

yes

PR #1496, opened 2021-03-21
Dynamically let a question set (or change) a baseline

gregelin:
**DRAFT WORK IN PROGRESS**

This PR allows a question to initially set or to subsequently change the baseline associated with the system.

Update `controls.models.Element.assign_baseline_controls` method to update controls to a different baseline.
    
Method `assign_baseline_controls` method now dynamically adds and removes controls from different baselines while leaving controls that exist in both baselines not changed.


gregelin:
Closing bc PR target branch is wrong. Target branch should not be main.

PR #1497, opened 2021-03-21
Dynamically let a question set (or change) a baseline 

gregelin:
Update controls.models.Element.assign_baseline_controls method
to update controls to a different baseline.

assign_baseline_controls method now dynamically adds and
removes controls from different baselines while leaving controls
that exist in both baselines not changed.


gregelin:
This branch should be very similar to `ge/question-adds-component`.

davidpofo (approved):
lgtm

PR #1499, opened 2021-03-23
Fix project page mini-dashboard displayed control count

gregelin:
Use `project.system.root_element.controls.all` to get count
of controls selected for a system instead of
`project.system.smts_control_implementation_as_dict`.

The former returns a constant value based on controls selected
and is the correct measure of controls assigned to the system.`


davidpofo (approved):
lgtm

PR #1500, opened 2021-03-23
Process question system actions to set project, system name

gregelin:
Add action processing for question actions targeted at system
to handle `system/update_system_and_project_name/<value>` to
set system name and project title.


davidpofo (approved):
lgtm

PR #1503, opened 2021-03-24
Release/0.9.3.1 fix

gregelin:
Fix Speedy SSP OSCAL JSON generation
    
    Speedy SSP OSCAL JSON generation was failing with the error:
    
    ```
    There was an error executing the template 'ssp_intro' output document 'ssp_v1_oscal_json' system-security-plan->system-implementation->users->0->title: sequence item 5: expected str instance, int found
    ```
    The error was happening in guidedmodules.module_logic render_output,
    and was caused by the template requiring a parameter from system_info_technical.yaml.
    
    The ssp_v1_oscal_json.md template was expecting to be able to get a value from
    system_info_technical.yaml.security_impact_users:
    
    ```
            "system-implementation": {
                "users": {
                    "%dict": "user in project.system_info_technical.security_impact_users",
                    "%value": {
                        "%key": "{{ oscal.make_uuid() }}",
                        "title": "{{ user.answer['Role'] }}"
                    }
                },
    ```
    
    This value was not forthcoming because system_info_technical.yaml had
    not been included in Speedy SSP app.yaml file. Including the file in app.yaml
    resolved the issue.
    
    A better solution in the future would be to gracefully handle missing parameter
    references and/or provide more precise feedback.


gregelin (commented):
Looks good.

PR #1505, opened 2021-03-30
assorted files changed to accommodate (future) branding

kerryrm:
some CSS changes, added a field to admin to input "site name" for use as a variable throughout the site. Added breaks to separate some forms and removed H3 tags around buttons. Also created "splash" as an include to allow branding without editing index.html. Footer is also now an include.


gregelin:
@kerryrm Most of the changes look good.

The core question is whether we should create a new model "Sitename" or if we can use the existing "Organization" model for the same purpose?

kerryrm:
Sitename is for using as a variable on pages that are branded (including more than just the org name). Consider something like "Organization GRC Tool"

gregelin:
As per Kerry: "Because the org name isn't ManTech GRC Tool, it's just ManTech. Organization is specific to the org. Sitename is specific to the product. See how a page says "Turbo ATO"? That's the sitename, not the organization."

PR #1507, opened 2021-04-01
Improve database connections

gregelin:
Set database connections to close after each transaction.
Set database transactions to be atomic by default.


davidpofo (commented):
does this change affect automated test completion? @gregelin

davidpofo (approved):
lgtm

PR #1508, opened 2021-04-01
temp solution for the navbar being too wide before collapsing to a hamburger.

kerryrm:
Navbar will collapse at 1102px instead of 768px. This keeps the navbar divs from stacking causing occlusion of the context bar between 768px and 1102px screen widths. / removed bg code from home-user


gregelin (approved):
Looks good.

PR #1511, opened 2021-04-02
changes to accommodate bg image and logic in case sitename is null

kerryrm:
Added logic for sitename var and css for bg image per greg.


kerryrm:
something is wrong. ignore.

PR #1517, opened 2021-04-06
Try diff SNYK command to ignore SNYK-PYTHON-CRYPTOGRAPHY-1022152

gregelin:
PR will tell SNYK to ignore the error SNYK-PYTHON-CRYPTOGRAPHY-1022152 until 4/30/2021 b/c there is not currently a fix available for the issue.


gregelin:
Fix made in #1518

PR #1520, opened 2021-04-07
Da/add projectv UI v2

davidpofo:
adding some rest start stuff and removing compliance app change ability as users shouldn’t change compliance app version, only the project version.


gregelin (approved):
Looks good

PR #1521, opened 2021-04-08
Update local-examples for venv, 0.9.3

gregelin:
Tweak local-examples Ubuntu files for installing with venv and using v0.9.3.x

This update is paired with https://github.com/GovReady/govready-q-docs/pull/38


local-examples/local-ubuntu-postgres-nginx-gunicorn-supervisor-http/install-os-packages.sh
```diff
@@ -17,12 +17,18 @@ apt-get update
...
-python3 python3-pip \
+python3 python3-pip python3-venv \
 python3-yaml \
 graphviz pandoc \
+gunicorn \
 language-pack-en-base language-pack-en
 
-# Upgrade pip to version 20.1+
-echo "Upgrading Python pip..."
-python3 -m pip install --upgrade pip
```
davidpofo, line 27: Does the pip upgrade happen implicitly now with the pip install on line 20?

PR #1522, opened 2021-04-08
Move project related modal inclusion to project-base.html

gregelin:
Project related modals are only used with pages that extend
the project-base.html template. Keeping the modal templates
on project-base.html reduces updates that need to made to
project.html branded pages which are more likely to change
in branding exercises than project-base.html page.


davidpofo (approved):
lgtm PR party 4/08

PR #1528, opened 2021-04-11
Initial refactor question actions into a new `action` question type

gregelin:
Begin refactoring to create a distinct `action` question type
to better handle actions associated with imputed questions.

Actions added to questions are only processed when the answer is saved
and gets a value. However, questions with imputed answers are not processed
by the `save_answer` method and do not have a `stored_value`. Therefore,
we can create a separate question type designed to check answers of other questions,
including imputed questions, and then process the answer and actions.

This distinct action type also allows us to optionally prompt user if they want us
to take the action.


gregelin:
Overtaken by events.

PR #1535, opened 2021-04-16
Updating move_project in views.py for Issue 1252

ztzerhouni:
1. Checks if the user is a superuser and can move the portfolio regardless of any other details.
2. Checks if the user is the owner/admin of the project and has edit access to the target portfolio.
3. Users continue to have access to a project after it moves. It is accessible from the project tab in the nav bar.
4. Each user who is a member of the project is given view access to the target portfolio.

Note: There is an issue with the except block in move_project. The following line caused an error after unsuccessful move is logged.

       return JsonResponse({ "status": "error", "message": sys.exc_info() })

### DIFF
--- a/siteapp/views.py
+++ b/siteapp/views.py
@@ -20,7 +20,7 @@ from django.utils import timezone
 from django.views.decorators.http import require_http_methods
 from django.views.generic import ListView
 from guardian.decorators import permission_required_or_403
-from guardian.shortcuts import get_perms_for_model
+from guardian.shortcuts import get_perms_for_model, get perms, assign_perm
 
 from controls.forms import ImportProjectForm
 from controls.views import add_selected_components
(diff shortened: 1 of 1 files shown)


siteapp/views.py
```diff
@@ -1370,15 +1370,24 @@ def move_project(request, project_id):
...
+        new_portfolio_id = request.POST.get("new_portfolio", "").strip()
         project = get_object_or_404(Project, id=int(project_id))
         cur_portfolio = project.portfolio
         new_portfolio = get_object_or_404(Portfolio, id=int(new_portfolio_id))
+    except:
+        return JsonResponse({"status": "error", "message": "Portfolio entered does not exist."})
+
+    # Check if the user moving the project is a superuser or
+    # if they are the owner of the project and have edit permissions in the target directory
+    if request.user.is_superuser or ((request.user in project.get_admins()) and 'change_portfolio' in get_perms(request.user, new_portfolio)):
```
davidpofo, line 1385:
```suggestion
    owner = True if request.user.has_perm('can_grant_portfolio_owner_permission', cur_portfolio) else False
    if request.user.is_superuser or ((request.user in project.get_admins()) or owner or 'change_portfolio' in get_perms(request.user, new_portfolio)):
```
@ztzerhouni 
You already check edit access with the change_portfolio and admin status with project.get_admins but you also need to check if they are an owner.
- [X] 1. Checks if the user is a superuser and can move the portfolio regardless of any other details.
- [ ] 2. Checks if the user is the owner/admin of the project and has edit access to the target portfolio.
To check if the user is an owner of the target portfolio this is an option.
`owner = True if request.user.has_perm('can_grant_portfolio_owner_permission', new_portfolio) else False`
- [X] 3. Users continue to have access to a project after it moves. It is accessible from the project tab in the nav bar.
- [X] 4. Each user who is a member of the project is given view access to the target portfolio.

siteapp/views.py
```diff
@@ -1370,15 +1370,25 @@ def move_project(request, project_id):
...
         project = get_object_or_404(Project, id=int(project_id))
         cur_portfolio = project.portfolio
         new_portfolio = get_object_or_404(Portfolio, id=int(new_portfolio_id))
+    except:
+        return JsonResponse({"status": "error", "message": "Portfolio entered does not exist."})
+
+    # Check if the user moving the project is a superuser or
+    # if they are the owner of the project and have edit permissions in the target directory
+    owner = True if request.user.has_perm('can_grant_portfolio_owner_permission', cur_portfolio) else False
+    if request.user.is_superuser or ((request.user in project.get_admins()) or owner and 'change_portfolio' in get_perms(request.user, new_portfolio)):
```
ztzerhouni, line 1386:
```suggestion
    if request.user.is_superuser or (request.user in project.get_admins() or owner and 'change_portfolio' in get_perms(request.user, new_portfolio)):
```
There was an extra set of parentheses around project.getadmins(). Removed the parentheses.

davidpofo (approved):
lgtm

PR #1536, opened 2021-04-17
Support portfolio selection after app selection

gregelin:
Prompt user for which portfolio to start project
in if portfolio value not previously set before visiting project templates.
Enables users to start a project from a blank link.

Update templates/portfolios/select-portfolio-modal.html to
get required information if started afterward and properly
target the the correction action


gregelin:
This PR is superseded by PR #1627 which removes the need to select a portfolio when starting a new project.

PR #1537, opened 2021-04-17
Add clear links for forgot password, change password

gregelin:
Added clearer links to take user to "forgot password" page and "change password" page.


templates/navbar.html
```diff
@@ -98,6 +98,7 @@
               </li>
               {% endif %}
               <li role="separator" class="divider"></li>
+              <li><a href="/accounts/password/change/">Change Password</a>{{environment.}}</li>
```
davidpofo, line 101: Is that period in environment supposed to be there? Also where does that variable come from? @gregelin

gregelin: I think `{{environment.}}` is accidental debug/dev code I failed to delete.

davidpofo: okay i'll delete it directly.

davidpofo (approved):
lgtm pr party 4/22

PR #1563, opened 2021-04-22
Immediately assign change project perms to user starting project

gregelin:
This fixes a bug where non-admin users were not assigned permissions
to change a project started by the user until after several project
modification steps were performed. The fix assigns non-admin project
change permissions right after project and system is created and
before further actions are made to project, such as setting baseline controls.

This is an import bug fix because non-admin users currently experience incorrect behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ v999 (April XX, 2021)
 
 **Bug fixes**
 
-* Replace me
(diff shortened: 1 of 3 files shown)


davidpofo (approved):
lgtm

davidpofo (approved):
lgtm pr party 4/22

PR #1565, opened 2021-04-22
Fix a few typos

peterkaminski:
Fixes three user-visible typos, and one in a comment.

The important one is changing "Checkout" (not a word) to "Check out" (proper English) on the home page (`home-user.html`).


gregelin (approved):
looks good

PR #1570, opened 2021-04-23
Faster retrieval of component control statements, show control titles

gregelin:
Retrieve component control statement much faster by getting
just related catalog_control_as_dict for statement. Also
add statement property to get control title. Display control title
on component control pages.

Important performance improvement!


davidpofo (approved):
lgtm

PR #1572, opened 2021-04-23
Am/feature inactivity timeout

azharem:
1. Better UX experience for when session ends due to inactivity

```
I want GovReady-Q to automatically navigate me to the login page instead of continuing to display in the browser window the previously loaded GovReady-Q page.

Timeout should probably be a system settings in the database.

I know this is done when:

The session times out due to inactivity
Activity is identified by no up or down movement or clicks within the page 
The time amount for timeout has to be an adjustable variable 
CMS settings 30 mins 
Once logged back in user redirects to  Home page 
```

2. Time Alert - Better UX experience for when session ends due to inactivity
```
As a GovReady user, when my session times out due to inactivity,
I want to receive an alert that my session was going to time out in 5 minutes.

So that I wont be logged out while still working on stuff. 

I know this is done when: 

The system will send a notification when there is a certain number of minutes of inactivity.  This number of minutes is configurable in environment deployment setting and by default is set to 20.
```


install.py
```diff
@@ -116,7 +116,10 @@ def create_environment_json(path):
     "static": "static_root",
     "secret-key": secret_key,
     "test_visible": False,
-    "debug": True
+    "debug": True,
+    "session_security_expire_at_browser_close" : True,
```
davidpofo, line 120: @alexanderward some more settings for the docker env.

siteapp/settings.py
```diff
@@ -403,6 +406,17 @@ def make_secret_key():
...
+# Session security and inactivity timeout. Logout user after certain period of inactivity.
+# By default user is warned at 20 minutes that session is about to expire and if user does not perform any mouse/keyboard activity
+# the session expires 10 minutes later (total of 30 minutes).
+warn_after = 85800
+expire_after = 86400
+expire_at_browser_close = True
+
+SESSION_EXPIRE_AT_BROWSER_CLOSE = environment['session_security_expire_at_browser_close'] if not DEBUG else expire_at_browser_close
+SESSION_SECURITY_WARN_AFTER = environment['session_security_warn_after'] if not DEBUG else warn_after
+SESSION_SECURITY_EXPIRE_AFTER = environment['session_security_expire_after'] if not DEBUG else expire_after
```
davidpofo, line 418:
```suggestion
SESSION_EXPIRE_AT_BROWSER_CLOSE = environment['session_security_expire_at_browser_close'] if not DEBUG else True
SESSION_SECURITY_WARN_AFTER = environment['session_security_warn_after'] if not DEBUG else 85800
SESSION_SECURITY_EXPIRE_AFTER = environment['session_security_expire_after'] if not DEBUG else 86400
```
Meant to add this on my last review. The 3 variables are not used anywhere so just applying them to the one-liner as they are defaults.

davidpofo:
@azharem Make sure to address the branch conflicts and you are go to go!

davidpofo (approved):
lgtm after I resolved the conflicts with develop

install.py
alexanderward: Can you add the same line to `dev_env/core/runner.py` under the `def generate_config` section?

PR #1573, opened 2021-04-24
Display control catalog guidance text in details tag next to component control implementation statements.

gregelin:
Display control catalog guidance text in `details` tag next to component control implementation statements.

Screen Shot 2021-04-24 at 10 24 28 AM


davidpofo (approved):
lgtm

PR #1575, opened 2021-04-24
Search component library by tag, make component tags clickable

gregelin:
Search component library by tag, make component tags clickable to improve search of catalog.


davidpofo (approved):
lgtm

PR #1576, opened 2021-04-24
Better notifification when project smt differs from certified

gregelin:
Better notify users when project implementation statement
differs from certified by displaying notice in third column control detail pages.

Improve language notifying users that project implementation statement
differs from certified. Only difference notice is clickable now.

Screen Shot 2021-04-24 at 10 17 29 AM


davidpofo (approved):
lgtm

PR #1579, opened 2021-04-25
Add input_type to AppInputs

gregelin:
Add input_type to AppInputs in order to make selections of input files based on type. This will support importing POA&Ms as part of loading and the starting an app.


davidpofo:
@gregelin implement somewhere. tests sound good too. PR party 5/20

davidpofo (approved):
pr party 5/27

PR #1581, opened 2021-04-26
[WIP] Replace branding with dynamic themes

gregelin:
Initial work to replace current branding with theme selection among multiple possible themes.

Create a "themes" directory for placing theme directories. Auto discover and register themes in settings.py.

Optional template directories are appended to the templates list to have lower precedence than GovReady's default template.

A new method `siteapp.views.get_site_theme` is used to select the appropriate template to render when theme value is set in `system_settings.Sitename`.

The set theme is passed into all templates via the `siteapp.context_processor.get_theme` where the `|add` filter can be used to dynamically set the path for extended or included templates.

To see theme switching work:
1. Open up http://localhost:8000/ and login so you are the `home-user.html` template from the `siteapp.views.home_user` view. 
2. Switch between branding themes in the `themes` directory by entering the name of the directory/theme in http://localhost:8000/admin/system_settings/sitename/1/change/ admin page.

The correct template is selected in the line `return render(request, [f"{get_site_theme()}home-user.html", "home-user.html"], {`

NOTE: the correct template is being called, but not the correct logos. That still has to be figured out.


gregelin:
Overtaken by events.

PR #1586, opened 2021-04-28
Release/v0.9.3.5

davidpofo:


templates/systems/components_selected.html
```diff
@@ -32,7 +32,7 @@
                             <select class="producer_element_id" id="producer_element_id" name="producer_element_id" onchange="$(form).submit();" style="width:340px;">
                                 <option value="None" selected disabled hidden>Select a Component</option>
                                 {% for component in elements %}
-                                    {% if component not in system_elements %}
+                                    {% if component not in system.producer_elements %}
```
gregelin, line 35: This could be a problem in the future. We should revisit what components get listed.

davidpofo: Not sure what that problem would be as this is system specific listing of components. the `system_elements` is actually the queryset of components that fit the pagination limit, while `system.producer_elements` is the full list of components.

gregelin: Going to approve this for now. But noting this to revisit.

PR #1589, opened 2021-05-03
Allow more Git URL styles in Add app source

mguelfi:
The current URL style check is a bit too tight and raises an unhandled exception if you use other valid Git URLS (like ssh://host:port/path/repository.git)

The check only seems to serve to select between HTTP(S) and SSH widgets, so this PR simplifies that to HTTP(S) and "other". Git itself gives good error messages that flow back into the form without needing to do any other handling.

There's also a bit of mixed whitespace clean-up done to bring admin.py into line w/ the rest of the code.


gregelin:
@mguelfi Howdy. Reviewing your PR now.

gregelin:
@mguelfi Could you please re-submit this PR against `develop` branch instead of `main`?

Our workflow is to pull changes into `develop` first, then do new releases into `main`.

mguelfi:
Sorry @gregelin , opened a new PR against develop.

PR #1591, opened 2021-05-11
Allow more Git URL styles in Add app source

mguelfi:
The current URL style check is a bit too tight and raises an unhandled exception if you use other valid Git URLS (like ssh://host:port/path/repository.git)

The check only seems to serve to select between HTTP(S) and SSH widgets, so this PR simplifies that to HTTP(S) and "other". Git itself gives good error messages that flow back into the form without needing to do any other handling.

There's also a bit of mixed whitespace clean-up done to bring admin.py into line w/ the rest of the code.


gregelin:
@mguelfi Thanks for resubmitting!

gregelin:
@mguelfi I did some testing of your submission last night. This particular file seems to have an odd mixing of tabs and spacing. So I'm still testing. Wanted to let you know.

mguelfi:
@gregelin I missed some tabs when I redid the cleanup.. sorry about that. New commit should be clean.

davidpofo (approved):
lgtm now

PR #1592, opened 2021-05-14
Fix multiple copies of same component displaying in component library search

gregelin:


gregelin:
Close and fix against updated develop

PR #1593, opened 2021-05-14
Fix multiple copies of same component displaying in component library search

gregelin:
Fixed by adding `distinct()` to search command.


gregelin:
Closing again

PR #1594, opened 2021-05-16
Fix cmpt search result, smt count on action, session timeout err

gregelin:
Fix various issues. Fix component search results returning multiple
copies. Fix Session Timeout showing 500 error after being away
from site. Fix control count on actions where too many control
statements were trying to be added instead of statements of
control implementation prototypes.


gregelin:
Merging bug fixes into `main` while upgrade to Django 3.1.9 is still being done in `develop`.

gregelin:
Superceded by #1595

PR #1598, opened 2021-05-20
Use self.element.tags.exists() in checking if tags exist.

gregelin:
Small change I forgot to commit. Use self.element.tags.exists().


davidpofo (approved):
PR 5/20 lgtm

PR #1603, opened 2021-05-27
Automatically clear, refresh output document content downloading docs

gregelin:
Performnce of document generation now sufficiently fast
to not require cache and manual "Refresh documents" button.


davidpofo (approved):
PR party 5/27

PR #1607, opened 2021-05-29
Ge/file upload extensions

gregelin:
Add tests for validating uppercase extensions on file uploads


gregelin:
Tested and passed...

PR #1610, opened 2021-06-02
Add 'Create a template' button to template library

gregelin:
Add "Create a template" button to template library so admins can create a new template (e.g., compliance app).


davidpofo (approved):
lgtm

PR #1612, opened 2021-06-03
'Back' link to question to take user to previous question

gregelin:
Add 'Back' link to question to take user to previous question.
Currently only supports taking end user back one previous question, not a que of previous questions.

Screen Shot 2021-06-09 at 6 09 02 AM

### DIFF
--- a/templates/question.html
+++ b/templates/question.html
@@ -527,14 +527,17 @@ Next Question: {{q.spec.title}} | {{task.title}}
                     {% endif %}
                     &raquo;
                   </button>
-
+                </div>
+                <div>
+                  {% if back_url %}
+                    <a href="{{back_url}}" id="back-button" class="btn btn-link question-button">previous question</button>
+                  {% endif %}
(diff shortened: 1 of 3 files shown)


templates/question.html
```diff
@@ -527,14 +527,17 @@ <h1>{{title|safe}}</h1>
                     {% endif %}
                     &raquo;
                   </button>
-
+                </div>
+                <div>
+                  {% if back_url %}
+                    <a href="{{back_url}}" id="back-button" class="btn btn-link question-button">previous question</button>
+                  {% endif %}
```
davidpofo, line 534:
```suggestion
                  {% if back_url %}
                      <div class="pull-left">
                          <button id="back-button" class="btn  btn-default btn-link question-button"><a href="{{back_url}}" style="color: inherit;">previous question</a></button>
                      </div>
                  {% endif %}
```

davidpofo (commented):
almost there.

templates/question.html
gregelin: @davidpofo I partially accepted suggestion. Moved back button to left but modified styling different than suggested.

PR #1613, opened 2021-06-03
Content-Security-Policy header permit images (*), videos youtube, vimeo

gregelin:
Adjust Content-Security-Policy header permit images from any source and videos from youtube, vimeo. This is to allow videos and pictures to be embedded in the questions.


davidpofo (approved):
lgtm

PR #1626, opened 2021-06-11
Ge/compare to select

gregelin:
Rename 'compare' column to 'select' in component library


gregelin:
Just a text label changing. Approving.

PR #1627, opened 2021-06-13
Remove portfolio selection modal from Start a Project process

gregelin:
Start projects in user's default portfolio to reduce the clicks
starting a project.

Use the User.create_default_portfolio_if_missing method consistently
to consistently create the user portfolio default portfolio.

Remove the PortfolioSignupForm because registration is no longer
used in registering users.

Remove the '"project_form": AddProjectForm(request.user ...' passed
into many templates because the navbar start app option no longer
brings up the portfolio select modal.


davidpofo (approved):
lgtm

PR #1629, opened 2021-06-16
Link to library version of component from a system's selected control component

gregelin:
Make it easy to navigate from a component listed as part of a control to the component in the library.


davidpofo (approved):
lgtm

PR #1633, opened 2021-06-17
try/except to still do the component search for non-Postgres users.

davidpofo:
If you try to search for a component without using a Postgres server an error occurs.


gregelin (approved):
looks great

PR #1634, opened 2021-06-19
Add a 'blank' project with no questions useful for batch project creation

gregelin:
Create a compliance app project template that has no questions and no output. This is useful because it allows maintaining of current workflow of starting projects, but creates a completely blank project which can then be adjusted to specific values.

Primary use of the blank project is to be used when batch creating dozens of projects.


gregelin:
This is just a data change -- no code is changed, just a new compliance app added to starter pack.

PR #1635, opened 2021-06-20
Avoid errors when project has no root_task set

gregelin:
We want to avoid errors and crashes when a project has no root task. Basic support for a project without a root task will allow us to occasionally create and manage empty projects.

Screen Shot 2021-06-20 at 7 10 26 AM


davidpofo (approved):
lgtm

PR #1636, opened 2021-06-20
Create OSCAL catalog for CMMC

gregelin:


davidpofo:
test is failing because test expects 3 default catalogs. need to change this to 4.  self.assertEqual(len(catalog_list), 3) -->  self.assertEqual(len(catalog_list), 4). Might want to check that the regex pattern change does not effect anything else.

davidpofo (approved):
lgtm

PR #1638, opened 2021-06-22
Add 'blank' compliance app to first_run

gregelin:
Make sure the 'blank' compliance app is loaded during the processing of first_run so that blank project is available in all installs.


davidpofo (approved):
lgtm

PR #1640, opened 2021-06-22
Auto-start a particular project

gregelin:
Enable the auto-starting of a particular project, e.g., start a project without having to visit the compliance app store.

Eventually, start a project and launch the first question.

Another way to think of this is one-click default project launch.


gregelin:
Needs to have a system setting value defined in the database so that you can optionally:
- [ ] indicate which project will auto be autostarted
- [ ] indicate if auto start project is on or off for the organization (maybe it is just one setting - if autoproject is set, then autostart

davidpofo (approved):
lgtm

alexanderward (approved):

PR #1648, opened 2021-07-02
Fix controls/0052 StatementTypeEnum migration

gregelin:
Correct controls/migrations/0052_auto_20210629_2328.py which had put the Python function in the wrong location.


gregelin:
Data migration fix. Tested locally.

PR #1650, opened 2021-07-06
Configure users on install

gregelin:
New govready_users parameter in local/environment.json to create sample users on install.

### DIFF
--- a/siteapp/management/commands/first_run.py
+++ b/siteapp/management/commands/first_run.py
@@ -123,6 +123,28 @@ class Command(BaseCommand):
                         username
                     ))
 
+        # Create default users, if specifified in local/environment.json
+        if len(settings.GOVREADY_USERS):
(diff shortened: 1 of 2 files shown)


siteapp/management/commands/first_run.py
```diff
@@ -123,6 +123,28 @@ def handle(self, *args, **options):
                         username
                     ))
 
+        # Create default users, if specifified in local/environment.json
+        if len(settings.GOVREADY_USERS):
+            for reg_user in settings.GOVREADY_USERS:
+                username = reg_user["username"]
+                if not User.objects.filter(username=username).exists():
+                    user = User.objects.create(username=username, is_superuser=True, is_staff=True)
```
davidpofo, line 131:
```suggestion
                    user = User.objects.create(username=username, is_superuser=False, is_staff=False)
```

davidpofo (approved):
lgtm

PR #1654, opened 2021-07-13
Da/dropnfill

davidpofo:
A second iteration on the component import which allows for drag and drop a file.


gregelin (approved):
looks good

PR #1658, opened 2021-07-16
Newline fix

alexanderward:


.gitattributes
```diff
@@ -0,0 +1,28 @@
+# Set the default behavior, in case people don't have core.autocrlf disabled.
+* text eol=lf
```
davidpofo, line 2: @alexanderward  Currently in our dockerfile we use dos2unix to handle CRLF issues. Could we just set EOL here to LF for all GovReady files? This would avoid all the times (and logs) produced when someone is starting up GovReady and/or when a container is run as part of a CICD pipeline.

davidpofo (approved):
lgtm

PR #1666, opened 2021-07-29
Import and display component, smts even if catalog missing

gregelin:
Import component control statement even if catalog not found
    
Remove test to see of control exists in control catalog when importing a component because we may import a component that has controls from a catalog that is not yet in GovReady.
    
The test on the control id being in the catalog was originally done to avoid importing a bad control (or bad control id). But it is worse to not be able to import the component because we don't yet have the a control catalog mentioned by the component.

Do not crash on displaying a component and component control implementation statement when a catalog associated with statements is missing from GovReady-Q install.


davidpofo (approved):
lgtm

PR #1675, opened 2021-08-05
Da/discussion updates

davidpofo:
### Unless otherwise discussed below the methods are already ready to be added given the Generic Relationship Discussion can have with any other model (given its GenericForeignKey)
- is_discussion_deleted() is now .is_discussion_inactive() (note there is no deleting of a discussion yet implemented.)
- get_project_context_dict doesn't seem relevant for any other model besides task.
- is_public() doesn't seems to do anything as assigning the attribute is_discussion_public to a model doesn't happen anywhere so it always returns the default which is False.
- Added get_discussion_autocompletes changes to get the organization from the discussion(should be 1 to 1). If another models wants to implement it they can look at the get_discussion_autocompletes in TaskAnswer
- on_discussion_comment assumes a task to get the org which is easy enough to replace (comment.discussion.organization) however it also needs a project for the task which you cannot get from the comment.
- on_discussion_comment_edited is not implemented see TODO line 95 in discussion/views.py
- get_user_role is implemented solely assuming a task in `TaskAnswer`

### DIFF
--- a/discussion/models.py
+++ b/discussion/models.py
@@ -176,6 +176,11 @@ class Discussion(models.Model):
     def is_public(self):
         return getattr(self.attached_to_obj, 'is_discussion_public', lambda : False)
 
+    def is_discussion_inactive(self):
+
+        if self.attached_to_obj is not None:
+            return True if self.attached_to_obj.title == "<Deleted Discussion>" else False
+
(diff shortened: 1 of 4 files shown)


gregelin (approved):
Looks good.

PR #1683, opened 2021-08-12
Fix first_run and add friendlier component import by refactoring source catalog handling

gregelin:
Import components and their statements even when catalog
not found or statement control ids are not found in referenced catalog.

Fix first_run. Move user creation earlier in first_run.
Update sample components to OSCAL 1.0.0.


controls/utilities.py
```diff
@@ -68,26 +68,28 @@ def oscalize_control_id(cl_id):
...
-def oscalize_catalog_key(catalogkey):
-    """ Covers empty catalog key case. Otherwise, outputs an oscal standard catalog key from various common formats for catalog keys
-    NIST_SP_800_53_rev4 --> NIST_SP-800-53_rev4
+def oscalize_catalog_key(catalogkey=None):
+    """Outputs an oscal standard catalog key from various common formats for catalog keys
+    Covers empty catalog key case
+    Example: NIST_SP_800_53_rev4 --> NIST_SP-800-53_rev4
     """
+    DEFAULT_CATALOG_KEY = 'NIST_SP-800-53_rev5'
+    if catalogkey is None or catalogkey=='':
```
alexanderward, line 77: Change:
if not catalogkey:
        return DEFAULT_CATALOG_KEY

alexanderward: tabs not showing up but you see what i  mean

controls/views.py
```diff
@@ -738,6 +738,16 @@ def statement_id_from_control(control_id, part_id):
 
         return f"{control_id}"
 
+    def generate_source(self, src_str):
+        """Return a valid catalog source given string"""
+        DEFAULT_SOURCE = "NIST_SP-800-53_rev5"
+        if src_str is None or src_str=='':
```
alexanderward, line 744: if not src_str:
            return DEFAULT_SOURCE

controls/views.py
```diff
@@ -1001,53 +996,42 @@ def create_control_implementation_statements(self, catalog_key, control_element,
...
-                )
-
-                logger.info(f"New statement with UUID {new_statement.uuid} created.")
-                statements_created.append(new_statement)
-
-            else:
-                logger.info(f"Control {control_id} doesn't exist in catalog {catalog_key}. Skipping Statement...")
-
+            control_id = implemented_control['control-id'] if 'control-id' in implemented_control else 'missing'
+            new_statement = Statement.objects.create(
```
alexanderward, line 1002: It's never good to do creates in a loop.  You are doing a SQL write per each item in a loop which is slow and taxing.  use Bulk create.
new_statements = []
for implemented_control in implemented_reqs:
  new_statement = Statement(sid=control_id, ....)
  new_statements.append(new_statement)
  
 # end of loop
statements_created =  Statement.objects.bulk_create(new_statements)

alexanderward (changes requested):

controls/utilities.py
gregelin: Fixed.

controls/views.py
gregelin: fixed.

controls/views.py
gregelin: bulk_created adopted.

alexanderward (approved):

PR #1698, opened 2021-08-24
Git configuration for EOL characters.

Tom-Camp:
Configuring Git using the steps outlined in this page: https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings


gregelin:
This PR may pending on adjusting `.gitattributes` to contain listings for all file types that should not be changed. File types of files that can be uploaded are listed in `discussion/validators.py`

gregelin (approved):

PR #1701, opened 2021-08-28
Prefix 'c' on 800-171 control ids for NIST OSCAL

gregelin:
Prefix 'c' on 800-171 control ids for NIST OSCAL.

Update catalog and baselines to prefix 'c' to control ids because NIST OSCAL requires ids to start with a letter because XML does.


gregelin:
Closing. Accidentally combined two requests.

PR #1703, opened 2021-08-29
Batch create CMMC component control content from existing content

gregelin:
Use a batch process to create component control implementation statements for CMMC catalog based on existing 800-53 control catalog component control implementation statements.

Create an admin command `makecmmcstatements` that accepts a list of component IDs and for each component looks at the existing statements. For each statement associated with a control ID that the CMMC catalog views as related, use that existing content to create a component control implementation statement for the related CMMC control.

This only creates control implementation statements for components in the library, not for component content in individual projects.

When creating the new CMMC statements associated with a component, also populate a new `RemoteStatement` model to track explicit "remote" control content relationships.

Also adding a JSONField to track a changelog for each change to a statement in order to eventually display a nice transaction log for a record.

- [x] Determine strategy & architecture
- [x] Figure out main loop
- [x] Add count of prototype statements to import record delete counts
- [x] De-oscalize controls to support ids like AC-20(2)
- [x] Make sure two controls contribute to the same control
- [x] Create controls.models.StatementRemote
- [x] Integrate StatementRemote
- [x] Add ChangelogJSONField to controls.models.Statement to have a complete history of changes
- [x] Write Statement method to append Changelog entry
- [x] Create, run migrations - uh, oh, will migration conflict with profile migration not yet merged?
- [ ] unit tests for migrated data
- [x] Revise loop
- [x] write management admin command
- [x] Show number of StatementRemotes that will be deleted
- [x] Extend to handle 800-171 catalog, too (also do with oscalizing 171 controls that don't have a letter )
- [x] Update CHANGELOG


controls/models.py
```diff
@@ -215,6 +217,29 @@ def _statement_id_from_control(control_id, part_id):
     def oscal_statement_id(self):
         return Statement._statement_id_from_control(self.sid, self.pid)
 
+    def change_log_add_entry(self, change):
+        # TODO: Test if entry is valid
+        if type(change) is not dict:
```
alexanderward, line 222: `if not isinstance(change, dict)`
is typically the proper way to do type checks in python

alexanderward (changes requested):

controls/models.py
gregelin: Change made.

alexanderward (approved):

PR #1709, opened 2021-09-09
Updated admin JSONField widgets to have a better display

alexanderward:


gregelin (approved):
Looks good. I added the widget to 'Catalog Data' and also fixed a model that needed `models.JSONField` instead of just `JSONField` for the widget.

PR #1726, opened 2021-09-22
Display inheritance information

gregelin:
Create `control.Inheritance` model and `control.Statement` foreign key to Inheritance to display inheritance information.


controls/models.py
```diff
@@ -240,6 +241,18 @@ class StatementRemote(auto_prefetch.Model):
     import_record = auto_prefetch.ForeignKey(ImportRecord, related_name="import_record_statement_remotes", on_delete=models.CASCADE,
                                              unique=False, blank=True, null=True, help_text="The Import Record which created this record.")
 
+class Inheritance(models.Model):
+    name = models.CharField(max_length=120, unique=True, blank=False, null=False)
+    description = models.TextField(unique=False, blank=True, null=True)
+    responsibility = models.TextField(unique=False, blank=True, null=True)
+
+    def __str__(self):
+        return "'%s'" % (self.name)
```
alexanderward, line 250: use F string:
`return f"{self.name}`

controls/models.py
```diff
@@ -240,6 +241,18 @@ class StatementRemote(auto_prefetch.Model):
...
+
+    def __repr__(self):
+        # For debugging.
+        return "'%s'" % (self.name, self.id)
```
alexanderward, line 254: use F String:
`return f"{self.name} - {self.id}"`

controls/views.py
```diff
@@ -1328,13 +1328,16 @@ def component_library_component(request, element_id):
         # Get the impl_smts contributed by this component to system
         impl_smts = element.statements_produced.filter(statement_type=StatementTypeEnum.CONTROL_IMPLEMENTATION_PROTOTYPE.name)
 
+    inheritances = Inheritance.objects.all()
```
alexanderward, line 1331: Should this always be all??  Or do we need a filter?

alexanderward (changes requested):

controls/views.py
gregelin: Yes. There are just three to five records in Inheritance, so really just using DB to allow content to change from install to install.

controls/models.py
gregelin: Updated.

controls/models.py
gregelin: Fixed

alexanderward (approved):

PR #1756, opened 2022-06-29
Upload systems by CSV and OSCAL

gregelin:


gregelin (commented):
We need to review ajax libraries. libraries should be bundled into GovReady bc it can be deployed in environments not connected to the Internet.

gregelin:
@SergioJFalcon I did some code refactoring.

PR #1767, opened 2022-07-25
Upgrade python libraries. Fix urlib3

gregelin:
Upgrade python libraries to fix urllib3 library issue


gregelin:
New PR with more info in #1768

PR #1796, opened 2023-02-05
Bulk update a component's statements across all systems

gregelin:
## Story
As an administrative user, I want to be able to propagate a change in a component's implementation statements to all all systems using that component.


gregelin:
Wrong branch targeted

PR #1812, opened 2023-06-11
Update Python packages

gregelin:
Update Python packages. Finally able to build again after release of urllib3 2.0.3.


gregelin:
Merging with issues into develop to be sure we have a version that is building. Then will fix other packages.

